PR #9, opened 2018-05-07
ENH: Measure Average

### LINKED ISSUE #5: measure_average
## Description

Average a number of `detectors` for `num` shots with a specified delay. The prior implementation had much of this logic written in the plan. However, this should user newer `bluesky` capabilities like:

* The `drop` message here: https://github.com/NSLS-II/bluesky/pull/928
* The `stream` callback: https://github.com/NSLS-II/bluesky/pull/932

```python
def measure_average(detectors, num=1, filters=None,
                    delay=None, drop_missing=True):
    """
    Gather a series of measurements from a list of detectors and return the
    average over the number of shots.
    
    Parameters
    ----------
    detectors : list
        List of detectors to measure at each event
    num : int, optional
        Number of samples to average together for each step along the scan
    delay : iterable or scalar, optional
        Time delay between successive readings
    Returns
    -------
    average : dict
        A dictionary of all the measurements taken from the supplied detectors
        averaged over `num` shots. In the event that a field is a string, or
        can not be averaged the last shot is returned
```

teddyrendahl:
## Description

A replacement for `meausure_average` with newer `bluesky` tech! Excited that the plan is so short! Here is what I added.

* `AverageStream` that absorbs a stream of raw data an re-emits it as an averaged stream. This is essentially a copy of what is contained in the `bluesky` tests with the exception that I removed any dependency on `streamz`.
* `measure_average` plan that subscribes a new `AverageStream` if one is not passed in, takes a measurement and returns the last event from the stream. 

### Thoughts
* The`AverageStream` caches its last_event on itself. This allows us to return this very succinctly at the end of the plan but feel slightly hacky?
* I created two new files. Feel free to suggest names for these if you don't like "plans" and "screens"

## Motivation and Context

Closes #5 

## How Has This Been Tested?

Tests added for both plan and stream

Docstrings

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,7 +1,7 @@
-pytest
 codecov
 doctr
+flake8
 ophyd
+pytest
 sphinx
 sphinx_rtd_theme
-flake8
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,6 +1,7 @@
 pytest
 codecov
 doctr
+ophyd
 sphinx
 sphinx_rtd_theme
 flake8
--- a/nabs/plan_stubs.py
+++ b/nabs/plan_stubs.py
@@ -1,0 +1,54 @@
+from bluesky.plans import count
+from bluesky.plan_stubs import subscribe
+from bluesky.preprocessors import stub_wrapper
+
+from nabs.streams import AverageStream
+
+
+def measure_average(detectors, num, delay=None, stream=None):
+    """
+    Measure an average over a number of shots from a set of detectors
+
+    Parameters
+    ----------
+    detectors : list
+        List of detectors to read
+
+    num : int
+        Number of shots to average together
+
+    delay: iterable or scalar, optional
+        Time delay between successive readings. See ``bluesky.count`` for more
+        details
+
+    stream : AverageStream, optional
+        If a plan will call :func:`.measure_average` multiple times, a single
+        ``AverageStream`` instance can be created and then passed in on each
+        call. This allows other callbacks to subscribe to the averaged data
+        stream. If no ``AverageStream`` is provided then one is created for the
+        purpose of this function.
+
+    Returns
+    -------
+    averaged_event : dict
+        A dictionary of all the measurements taken from the list of detectors
+        averaged for ``num`` shots. The keys follow the same naming convention
+        as that will appear in the event documents i.e "{name}_{field}"
+
+    Notes
+    -----
+    The returned average dictionary will only contain keys for 'number' or
+    'array' fields. Field types that can not be averaged such as 'string' will
+    be ignored, do not expect them in the output.
+    """
+    # Create a stream and subscribe if not given one
+    if not stream:
+        stream = AverageStream(n=num)
+        yield from subscribe('all', stream)
+    # Ensure we sync our stream with request if using a prior one
+    else:
+        stream.n = num
+    # Measure our detectors
+    yield from stub_wrapper(count(detectors, num=num, delay=delay))
+    # Return the measured average as a dictionary for use in adaptive plans
+    return stream.last_event['data']
--- a/nabs/streams.py
+++ b/nabs/streams.py
@@ -1,0 +1,77 @@
+"""Standard data streams for automated routines"""
+import streamz
+import numpy as np
+from bluesky.callbacks.stream import LiveDispatcher
+
+
+class AverageStream(LiveDispatcher):
+    """
+    Stream that averages data points together
+
+    As event documents are emitted from the RunEngine they are collected by
+    AverageStream, averaged and re-emitted as a secondary event stream. This
+    allows callbacks to subscribe to the average stream for more elegant
+    visualization and analysis.
+
+    The average stream can either be constructed with a specific number of
+    points to average or this information can be passed through in the start
+    document metadata. See the :meth:`.start` for more information.
+
+    Parameters
+    ----------
+    n : int, optional
+        Number of points to average together
+    """
+    def __init__(self, n=None):
+        self.n = n
+        self.in_node = None
+        self.out_node = None
+        self.averager = None
+        super().__init__()
+
+    def start(self, doc):
+        """
+        Create the stream after seeing the start document
+
+        The callback looks for the 'average' key in the start document to
+        configure itself.
+        """
+        # Grab the average key
+        self.n = doc.get('average', self.n)
+        # Define our nodes
+        if not self.in_node:
+            self.in_node = streamz.Source(stream_name='Input')
+
+        self.averager = self.in_node.partition(self.n)
+
+        def average_events(cache):
+            average_evt = dict()
+            desc_id = cache[0]['descriptor']
+            # Check that all of our events came from the same configuration
+            if not all([desc_id == evt['descriptor'] for evt in cache]):
+                raise Exception('The events in this bundle are from '
+                                'different configurations!')
+            # Use the last descriptor to avoid strings and objects
+            data_keys = self.raw_descriptors[desc_id]['data_keys']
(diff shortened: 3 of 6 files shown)


dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
-pytest
 codecov
 doctr
+flake8
+ophyd
+pytest
```
ZLLentz: I appreciate the alphabetization here

nabs/plan_stubs.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    # Create a stream and subscribe if not given one
+    if not stream:
+        stream = AverageStream(n=num)
+        yield from subscribe('all', stream)
+    # Ensure we sync our stream with request if using a prior one
+    else:
+        stream.n = num
+    # Measure our detectors
+    yield from stub_wrapper(count(detectors, num=num, delay=delay))
```
ZLLentz: Does this just strip the run_wrapper off of count?

nabs/streams.py
```diff
@@ -0,0 +1,73 @@
...
+    points to average or this information can be passed through in the start
+    document metadata. See the :meth:`.start` for more information.
+
+    Parameters
+    ----------
+    n : int, optional
+        Number of points to average together
+    """
+    def __init__(self, n=None):
+        self.n = n
```
ZLLentz: Can you make this attribute a searchable key? e.g. not 1 character long

nabs/streams.py
```diff
@@ -0,0 +1,73 @@
...
+        self.last_event = None
+        self.raw_cache = list()
+        super().__init__()
```
ZLLentz: Normally we do this first in an init. Does it matter for `LiveDispatcher` subclasses to do this last? Asking for knowledge gain.

nabs/plan_stubs.py
```diff
@@ -0,0 +1,54 @@
...
+    Notes
+    -----
+    The returned average dictionary will only contain keys for 'number' or
+    'array' fields. Field types that can not be averaged such as 'string' will
+    be ignored, do not expect them in the output.
+    """
+    # Create a stream and subscribe if not given one
+    if not stream:
+        stream = AverageStream(n=num)
+        yield from subscribe('all', stream)
```
ZLLentz: btw, does this need your bluesky PR to work?

teddyrendahl:
This requires https://github.com/NSLS-II/bluesky/pull/1022

dev-requirements.txt
teddyrendahl: 👍

nabs/plan_stubs.py
teddyrendahl: Correct!

nabs/streams.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def start(self, doc):
+        """
+        Create the stream after seeing the start document
+
+        The callback looks for the 'average' key in the start document to
+        configure itself.
+        """
+        # Grab the average key
+        self.n = doc.get('average', self.n)
```
ZLLentz: This is an interesting idea, I haven't seen any code take action based on start doc metadata before. This doesn't appear usable in `measure_average` because num is a required argument, but I'm assuming the point is to leave this here for use later?

nabs/streams.py
teddyrendahl: Shouldn't matter. I'll add it to top for consistency

nabs/plan_stubs.py
teddyrendahl: Was late posting below. Yes it does.

nabs/streams.py
teddyrendahl: This was a suggestion from the NSLS team when we were there. I liked the idea and left it in though I think taking it to the extreme could be scary. The crucial thing that could break this plan is that the callback is out of sync configuration-wise, giving multiple ways to sync things seems reasonable.

nabs/streams.py
```diff
@@ -0,0 +1,73 @@
...
+        self.raw_cache.append(doc)
+        # If we have enough events average
+        if len(self.raw_cache) == self.n:
+            average_evt = dict()
+            # Check that all of our events came from the same configuration
+            desc_id = self.raw_cache[0]['descriptor']
+            if not all([desc_id == evt['descriptor']
+                        for evt in self.raw_cache]):
+                raise Exception('The events in this bundle are from '
+                                'different configurations!')
```
ZLLentz: This exception shouldn't happen in a normal run, but it can probably happen if we subscribe `AverageStream` to another `LiveDispatcher` that is a bit malformed, so I like the inclusion. The `RunEngine` should throw it's own exception that will prevent us from reaching this one during normal use.

nabs/streams.py
```diff
@@ -0,0 +1,73 @@
...
+            # Use the last descriptor to avoid strings and objects
+            data_keys = self.raw_descriptors[desc_id]['data_keys']
+            for key, info in data_keys.items():
+                # Information from non-number fields is dropped
+                if info['dtype'] in ('number', 'array'):
```
ZLLentz: I like this approach a lot, no need to ask what to average, just average everything that you can.

nabs/streams.py
teddyrendahl: It can happen in a malformed `plan` as well.

```python
def bad_plan(
     # Wait for 10 shot average
     yield from subscribe(AverageStream(num=10))
     # Configure
     yield from configure(det, ...)
     # Only take one shot
     yield from trigger_and_read([det], ...)
     # Reconfigure
     yield from configure(det, ...)
```

nabs/streams.py
ZLLentz: no preference

nabs/streams.py
```diff
@@ -0,0 +1,73 @@
...
+                    # Average together
+                    average_evt[key] = np.mean([evt['data'][key]
+                                                for evt in self.raw_cache],
+                                               axis=0)
+            self.last_event = {'data': average_evt, 'descriptor': desc_id}
```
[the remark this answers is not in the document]
teddyrendahl: Nice catch. `process_event` sets a new `timestamp` for the event that just corresponds to the time the calculation was done, as well as pointing to the new descriptor e.t.c. So to answer your question, the `last_event` is malformed until it is put through `process_event`.

Do you think we should return the whole event here? The old `measure_average` only cared about what was contained within "data"  so I didn't care about the other stuff as much. I figured if you wanted full events you would just `subscribe`. Maybe this was an oversight.

ZLLentz:
I like this a lot. I'm still chewing on `last_event` vs subscribing a basic callback. `last_event` looks hacky, but in the end it's way more usable than having everyone who wants a really basic average make a collector, subscribe it, and check what happened. Still chewing, but I'll hit approve once the `n` attr is renamed and you can merge when the travis passes (e.g. your PR is accepted at `bluesky`)

tests/test_plans.py
```diff
@@ -0,0 +1,28 @@
+import logging
```
ZLLentz: Can you rename this file to `test_plan_stubs`?

nabs/streams.py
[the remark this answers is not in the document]
teddyrendahl: So just store the `"data"` as `last_average`?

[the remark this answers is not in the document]
teddyrendahl: Kind of leaning towards this. Clear delineation -> want context like descriptors and timestamps, multiple events you must subscribe. Want purely processed data for adaptive scans, grab it from the callback itself.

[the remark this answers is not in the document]
ZLLentz: Yeah, I like that, since the descriptor id doesn't mean anything anyway unless you're subscribing and get the descriptor doc, so as is it just adds an extra key to the workflow.

PR #10, opened 2018-05-15
ENH: Filter and drop utilities

ZLLentz:
## Description

- `drop_wrapper(plan, filters, max_dt)` that will do plan, but replace `save` with `drop` if the filters say so, or if the time between first and last read in an event bundle is greater than `max_dt`. This is a "cheap" way to check we were suspended to avoid desynced event documents.
- `count_events(detectors, num, delay)` that is literally count, but if `drop_wrapper` happened it repeats the measurement until `num` event documents have been emitted.

## Motivation and Context

Avoid re-implementing filter logic in every level. Utilize `drop` for great good. Abandoned the idea of trusting a `preprocessor` to cache and replay the "bad" reads, instead encapsulate this in a plan.

This means a plan like `measure_average` can use `count_events` and trust that if "someone" wants to filter at a higher level, we repeat reads that came up with bad values.

## How Has This Been Tested?

Added tests to high coverage.

We need to make the sphinx pages for this soon, but I will defer for this PR.

### DIFF
--- a/nabs/preprocessors.py
+++ b/nabs/preprocessors.py
@@ -1,3 +1,4 @@
+import collections
 import logging
 import time
 
@@ -24,6 +25,9 @@ class DropWrapper:
         read to after the last read is greater than this number.
     """
     def __init__(self, filters=None, max_dt=None):
+        if filters is not None:
+            if not isinstance(filters, collections.Iterable):
+                filters = [filters]
         self.filters = filters
         self.max_dt = max_dt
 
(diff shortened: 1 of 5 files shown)


nabs/preprocessors.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class DropWrapper:
+    """
+    Replaces ``save`` messages with ``drop`` if the event is bad.
+
+    Parameters
+    ----------
+    filters: ``dict``, optional
+        A dictionary mapping from read key to function of one argument. This
```
teddyrendahl: Either I'm confused or the docstring is wrong? Looks like you are expecting a list? I would recommend a `dict` so you can name your filters. Will make it more clear with multiple installed which is causing the drop.

nabs/preprocessors.py
```diff
@@ -0,0 +1,94 @@
...
+        self._in_filter = True
+        dt = self.last_read_time - self.first_read_time
+        if self.max_dt is not None and dt > self.max_dt:
+            logger.info(('Event took %ss to bundle, readings are desynced. '
+                         'Dropping'), dt)
+            return (yield from drop())
+        elif self.filters is not None:
+            for filt in self.filters:
+                if not filt(self.ret):
+                    logger.info('Bad event=%s. Dropping', self.ret)
```
teddyrendahl: This is where that `filter` name would be handy.

nabs/preprocessors.py
```diff
@@ -0,0 +1,94 @@
...
+    Replaces ``save`` messages with ``drop`` if the event is bad.
+
+    Parameters
+    ----------
+    filters: ``dict``, optional
+        A dictionary mapping from read key to function of one argument. This
+        is an "is_bad_value(value)" function that should return ``True`` if the
+        value is bad.
+
+    max_dt: ``float``, optional
```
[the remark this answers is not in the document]
ZLLentz: Noted, I'll change it. I agree that "dt" is not clear enough.

nabs/preprocessors.py
ZLLentz: Oops, didn't update the docstring here. My mistake. I agree that naming the filters could be useful, I will make this change and synchronize the docstrings.

ZLLentz:

PR #11, opened 2018-05-16
WIP: EventBuilder readable device

ZLLentz:
## Description

`EventBuilder(<signals>, <rate>)`
`EventBuilder.trigger()`
`EventBuilder.read()`
- maximize chance of matched timestamps via monitoring
- read names match native signal reads
- bin based on the rate (e.g. rate=100 -> bin on 1/100 of a second)
- stop when we have one full event

## Motivation and Context

More synchronized readings mean we can use more interesting filters, and we can correlate/normalize diode readings.

It's not clear to me if this even should be here, or if it should be something more fundamental with the RunEngine to ensure synchronous or near-synchronous readings. It may also be better as a flyer/event-builder stream scheme.

This can also be expanded to do the `FlyerInterface` and produce a number of consecutive timestamp-matched events.

## How Has This Been Tested?

WIP

WIP

### DIFF
--- a/nabs/event.py
+++ b/nabs/event.py
@@ -2,27 +2,31 @@ from collections import defaultdict
 from functools import partial
 from threading import Event, Thread
 import math
+import time
 
 from ophyd.status import Status
 
 
 class EventBuilder:
-    def __init__(self, signals, rate, duplicates='closest'):
+    def __init__(self, signals, rate, timeout=10.0, duplicates='closest'):
         self.signals = signals
         self.rate = rate
+        self.timeout = timeout
         self.duplicates = duplicates
         self.cbid = {}
+        self.start_ts = 0
         self._clear_events()
 
     def add(self, name, *, value, timestamp, **kwargs):
-        bin_ts = self._nearest_ts(timestamp)
-        delta = timestamp - bin_ts
-        event = self.bins[bin_ts]
-        event[name].append({'value': value,
-                            'timestamp': timestamp,
-                            'delta': delta})
-        if len(event) == len(self.signals):
-            self._has_evt.set()
+        if timestamp > self.start_ts:
+            bin_ts = self._nearest_ts(timestamp)
+            delta = timestamp - bin_ts
+            event = self.bins[bin_ts]
+            event[name].append({'value': value,
+                                'timestamp': timestamp,
+                                'delta': delta})
+            if len(event) == len(self.signals):
+                self._has_evt.set()
 
     def _nearest_ts(self, timestamp):
         expanded = timestamp * self.rate
@@ -35,6 +39,7 @@ class EventBuilder:
 
     def _monitor_events(self):
         if not self.cbid:
+            self.start_ts = time.time()
             for sig in self.signals:
                 cbid = sig.subscribe(partial(self.add, sig.name))
                 self.cbid[sig.name] = cbid
@@ -58,9 +63,9 @@ class EventBuilder:
         return status
 
     def _wait_trigger(self, status):
-        self._has_evt.wait()
+        success = self._has_evt.wait(timeout=self.timeout)
         self._stop_events()
-        status._finished(success=True)
+        status._finished(success=success)
 
     def read(self):
         timestamps = sorted(self.bins.keys())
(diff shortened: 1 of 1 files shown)


nabs/event.py
```diff
@@ -0,0 +1,109 @@
...
+    def _stop_events(self):
+        if self.cbid:
+            for sig in self.signals:
+                sig.unsubscribe(self.cbid[sig.name])
+        self.cbid = {}
+
+    def _clear_events(self):
+        self._stop_events()
+        self._has_evt = Event()
+        self.bins = defaultdict(partial(defaultdict, list))
```
teddyrendahl: This line confuses me. I question whether it is cleaner than just having a `get(key, default)` call...

nabs/event.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def trigger(self):
+        self._clear_events()
+        self._monitor_events()
+        status = Status(obj=self)
+        Thread(target=self._wait_trigger, args=(status,)).start()
+        return status
+
+    def _wait_trigger(self, status):
+        success = self._has_evt.wait(timeout=self.timeout)
```
teddyrendahl: So `trigger` will wait and then `read` will return immediately? For some reason I would have written it inverted but I can't think of a way to justify this 😄. Any precedent elsewhere, i.e `AreaDetector`?

teddyrendahl: Second thought. This is probably the correct way.

nabs/event.py
```diff
@@ -0,0 +1,109 @@
...
+from functools import partial
+from threading import Event, Thread
+import math
+import time
+
+from ophyd.status import Status
+
+
+class EventBuilder:
+    def __init__(self, signals, rate, timeout=10.0, duplicates='closest'):
```
teddyrendahl: `Duplicates` should be an Enum? Also maybe a class attribute? 🤷‍♂️

nabs/event.py
```diff
@@ -0,0 +1,109 @@
...
+                    elif self.duplicates == 'last':
+                        index = -1
+                    elif self.duplicates == 'closest':
+                        index = 0
+                        delta = math.inf
+                        for i, dct in enumerate(data):
+                            abs_del = abs(dct['delta'])
+                            if abs_del < delta:
+                                delta = abs_del
+                                index = i
```
teddyrendahl: `sorted(data, key=lambda dct: dct['delta'])[0]` will do this in one line unless you wanted to be very explicit about how you break ties?

nabs/event.py
ZLLentz: I agree that this line is confusing, but the alternative is also bad.
```
event = self.bins.get(ts, {})
data = event.get(name, [])
data.append(<datagram>)
event[name] = data
self.bins[ts] = event
```
The `defaultdict` lets you skip the reassigning part (e.g. omit the last two lines). If you think this is ok I can look into doing the reassigning everywhere, it might only be in 2 or 3 places.

nabs/event.py
ZLLentz: Yeah, this is how the synthetic devices work, so I just copied that pattern. I think this is preferred in general because you wait for triggers to complete in parallel, and then do the reads in series. Of course the distinction between `trigger` and `read` is totally lost if you only do it on one device.

nabs/event.py
ZLLentz: Probably correct, if only so that the choices available are clear.

nabs/event.py
ZLLentz: Hey man, you can't ding me for confusing lines in `defaultdict` and then suggest a one-liner here. Pick your team.

ZLLentz: I always forget the `sorted` `key=` api. This is probably the more pythonic approach, though it does do unnecessary work because I only need one element sorted.

nabs/event.py
teddyrendahl: Eh. Think its fine.

nabs/event.py
teddyrendahl: Fair enough!

ZLLentz:
Agree that this API is sub-optimal and your suggested API is better. I'll let this sit for a bit and we can talk about what the "best" API would be and then think about how to implement it. I will tackle updates based on your comments when this is revisited, probably after the ophyd 1.2.0 alpha release.

ZLLentz:

PR #12, opened 2018-05-16
WIP: Per-message Suspenders

ZLLentz:
## Description

Re-imagining of suspenders as a preprocessor. Seems to do what I intend but definitely still in the experimental phase. This will monitor a signal and halt the plan on particular messages until the signal reaches a good value.

## Motivation and Context

- Finer control over when to suspend (only on reads? only on moves?)
- No chance of multi-suspensions

## How Has This Been Tested?

WIP

WIP

### DIFF
--- a/nabs/suspenders.py
+++ b/nabs/suspenders.py
@@ -1,4 +1,5 @@
 import asyncio
+import copy
 import logging
 from threading import RLock, Event
 
@@ -120,17 +121,19 @@ class SuspendPreprocessor:
         with self._rlock:
             if self._cmd is None or msg.command in self._cmd:
                 if not self._ok_future.done() and not self._suspend_active:
-                    logger.debug('saw msg_proc(%s), suspend now', msg)
-
-                    def new_gen():
-                        self._suspend_active = True
-                        yield from wait_for([self._ok_future])
-                        self._suspend_active = False
-                        logger.info('Resuming plan')
-                        yield msg
-
-                    return new_gen(), None
+                    return self._suspend(msg), None
             return None, None
+
+    def _suspend(self, msg):
+        logger.debug('suspend on msg=%s', msg)
+        # Set this flag so we don't recursively check to suspend on wait_for
+        self._suspend_active = True
+        yield from wait_for([self._ok_future])
+        self._suspend_active = False
+        logger.info('Resuming plan')
+        # Yield a copy of the message so we can suspend again
+        # Bluesky skips preprocessor if it already saw the message
+        return (yield copy.copy(msg))
 
 
 class BeamSuspender(SuspendPreprocessor):
--- a/nabs/suspenders.py
+++ b/nabs/suspenders.py
@@ -130,7 +130,7 @@ class SuspendPreprocessor:
         self._suspend_active = True
         yield from wait_for([self._ok_future])
         self._suspend_active = False
-        logger.info('Resuming plan')
+        logger.debug('resuming plan in _suspend')
         # Yield a copy of the message so we can suspend again
         # Bluesky skips preprocessor if it already saw the message
         return (yield copy.copy(msg))
--- a/nabs/suspenders.py
+++ b/nabs/suspenders.py
@@ -1,0 +1,176 @@
+import asyncio
+import logging
+from threading import RLock, Event
+
+from bluesky.plan_stubs import null, wait_for
+from bluesky.preprocessors import plan_mutator
+
+logger = logging.getLogger(__name__)
+
+
+class SuspendPreprocessor:
+    """
+    Base preprocessor that suspends on specific commands based on a signal.
+
+    Parameters
+    ----------
+    signal: ``Signal``
+        The signal to subscribe to, whose value determines when to suspend.
+
+    commands: ``list of str``, optional
+        The commands to suspend on. If omitted, we'll suspend on all commands.
+
+    sleep: ``int`` or ``float``, optional
+        The amount of time to wait after `should_resume` returns ``True``
+        before ending the suspension. If `should_suspend` return ``True`` at
+        any time during this wait period, we will cancel the resumption and
+        return to the suspended state.
+    """
+    def __init__(self, signal, *, commands=None, sleep=0,
+                 pre_plan=null, post_plan=null, follow_plan=null):
+        self._sig = signal
+        self._cmd = commands
+        self._sleep = sleep
+        self._suspend_active = False
+        self._resume_ts = None
+        self._suspend_ev = Event()
+        self._ok_future = asyncio.Future()
+        self._ok_future.set_result('ok')
+        self._rlock = RLock()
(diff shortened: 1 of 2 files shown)


nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+    ----------
+    signal: ``Signal``
+        The signal to subscribe to, whose value determines when to suspend.
+
+    commands: ``list of str``, optional
+        The commands to suspend on. If omitted, we'll suspend on all commands.
+
+    sleep: ``int`` or ``float``, optional
+        The amount of time to wait after `should_resume` returns ``True``
+        before ending the suspension. If `should_suspend` return ``True`` at
```
teddyrendahl: returns

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+        any time during this wait period, we will cancel the resumption and
```
teddyrendahl: resumption ... Wow that is a word.

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Base preprocessor that suspends on specific commands based on a signal.
+
+    Parameters
+    ----------
+    signal: ``Signal``
+        The signal to subscribe to, whose value determines when to suspend.
+
+    commands: ``list of str``, optional
+        The commands to suspend on. If omitted, we'll suspend on all commands.
```
teddyrendahl: Is there commands we never want to `suspend` on? For instance I think we are always find "saving"? We are always fine "dropping"?

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+        self._resume_ts = None
+        self._suspend_ev = Event()
+        self._ok_future = asyncio.Future()
+        self._ok_future.set_result('ok')
+        self._rlock = RLock()
+        self._subid = None
+
+    def should_suspend(self, value):
+        """
+        Returns ``True`` if we should suspend.
```
teddyrendahl: Should document that this has to be implemented in children.

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+        Parameters
+        ----------
+        value: signal value
+            The value reported by a signal callback.
+        """
+        raise NotImplementedError()
+
+    def should_resume(self, value):
+        """
+        Returns ``True`` if we should resume.
```
teddyrendahl: I would note that the default behavior is just the reciprocal of the `should_suspend` unless implemented in child.

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+        The amount of time to wait after `should_resume` returns ``True``
+        before ending the suspension. If `should_suspend` return ``True`` at
+        any time during this wait period, we will cancel the resumption and
+        return to the suspended state.
+    """
+    def __init__(self, signal, *, commands=None, sleep=0,
+                 pre_plan=null, post_plan=null, follow_plan=null):
+        self._sig = signal
+        self._cmd = commands
+        self._sleep = sleep
```
teddyrendahl: Why are these private?

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+        The commands to suspend on. If omitted, we'll suspend on all commands.
+
+    sleep: ``int`` or ``float``, optional
+        The amount of time to wait after `should_resume` returns ``True``
+        before ending the suspension. If `should_suspend` return ``True`` at
+        any time during this wait period, we will cancel the resumption and
+        return to the suspended state.
+    """
+    def __init__(self, signal, *, commands=None, sleep=0,
+                 pre_plan=null, post_plan=null, follow_plan=null):
```
teddyrendahl: These plans are just dropped on the floor. Why have them as args?

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+                                self._sig.name, value)
+                    loop = asyncio.get_event_loop()
+                    self._ok_future = loop.create_future()
+                    self._suspend_ev.set()
+
+    def _run_release_thread(self):
+        logger.info('%s suspension is over, waiting for %ss then resuming.',
+                    self._sig.name, self._sleep)
+        loop = asyncio.get_event_loop()
+        loop.run_in_executor(None, self._release_thread)
```
teddyrendahl: Should this all clear any prior release threads? Imagining the following situation.

1. Suspender fires
2. Suspender returns back to operating region
2. Create release thread that waits for 5 seconds
3. Supsender fires
4. Suspender returns back to operation region, spawns a new 5 second wait
3. Release thread from first suspension resolves in less that 5 seconds after second suspension

nabs/suspenders.py
```diff
@@ -0,0 +1,179 @@
...
+        self._suspend_active = True
+        yield from wait_for([self._ok_future])
+        self._suspend_active = False
+        logger.debug('resuming plan in _suspend')
+        # Yield a copy of the message so we can suspend again
+        # Bluesky skips preprocessor if it already saw the message
+        return (yield copy.copy(msg))
+
+
+class BeamSuspender(SuspendPreprocessor):
```
teddyrendahl: This is awesome! Very clear and elegant. Means you did nice work in the class design 😉.

On a completely unrelated note it makes me want a global `BeamStats`.

nabs/suspenders.py
ZLLentz: You are probably correct, we can omit some commands by default, e.g. any commands that don't interact with devices.

nabs/suspenders.py
ZLLentz: Will-do

nabs/suspenders.py
ZLLentz: Will-do

nabs/suspenders.py
ZLLentz: I don't remember. I probably was worried about a malicious user changing them during a run. I can swap them to public if you think adjusting these is useful between runs.

nabs/suspenders.py
ZLLentz: I forgot to remove the args after I decided that these weren't useful.

nabs/suspenders.py
ZLLentz: Agreed on global `BeamStats`. This can be revisited.

ZLLentz: One caveat: we will have two beams in a year or more.

nabs/suspenders.py
teddyrendahl: I think the `RE` will grab the main thread so not worried about it. I guess they could do it on a `pause`? Not a big enough concern for me. I think they should be public.

nabs/suspenders.py
teddyrendahl: Sounds good. Thanks for the clarification.

ZLLentz:

PR #15, opened 2018-07-03
ENH: Golden Section Search

teddyrendahl:
## Description

This lays the groundwork for optimization methods using "Golden Section Search". This will be just one of the optimization methods that we eventually have available for operators.

### Features
* `golden_section_search` that finds the position of the minimum within a specific range.
* `optimize` function that flips a signal around if we want to maximize. Also, sets `md`, `limits`, 
     e.t.c. This is the major "switchyard" function.
* `minimize` and `maximize` which are just wrappers for optimize
* `walk_to_target` where we make an `ErrorSignal` object (just the absolute distance from our 
    target) and minimize this.

## How Has This Been Tested?

Tests for everything!

In progress!

### DIFF
--- a/nabs/optimize.py
+++ b/nabs/optimize.py
@@ -35,6 +35,38 @@ def maximize(*args, **kwargs):
 
 
 def walk_to_target(signal, motor, target, tolerance, **kwargs):
+    """
+    Walk the motor until a signal reaches our target
+
+    Similar to the :func:`.maximize` and :func:`.minimize`. There are options
+    for multiple algorithms to dictate the scanning procedure. This may change
+    the interpretation of values passed into this scanning procedure.
+
+    Parameters
+    ----------
+    signal: ophyd.Signal
+        Signal to maximize
+
+    motor: ophyd.OphydObject
+        Any set-able object
+
+    tolerance: float, optional
+        The tolerance in which our motor is
+
+    average: int, optional
+        Choice to take an average of points at each point in the scan
+
+    limits: tuple, optional
+        Limit the region the scan will search within. If this is not provided,
+        the soft limits of the signal will be used. In this case, these must be
+        configured or the scan will not be allowed to continue.
+
+    method: str, optional
+        Choice of optimization methods
+
+    md: dict, optional
+        metadata
+    """
     # Add walk information to metadata
     _md = {'plan_name': 'walk_to_target',
            'target': target}
@@ -99,6 +131,11 @@ def optimize(signal, motor, tolerance,
         else:
             raise ValueError("No limits provided or set on motor")
 
+    # Create an inverted signal if we need to maximize
+    if maximize:
+        raw = signal
+        signal = InvertedSignal(raw)
+
     # Create plan metadata
     _md = {'detectors': [signal],
            'motors': [motor],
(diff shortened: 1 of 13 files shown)


nabs/optimize.py
```diff
@@ -0,0 +1,289 @@
...
+    # Create a signal whose value is the absolute error
+    error = ErrorSignal(signal, target)
+    return (yield from minimize(error, motor, tolerance, **kwargs))
+
+
+def optimize(signal, motor, tolerance,
+             average=None, limits=None, method='golden',
+             maximize=False, md=None):
+    """
+    The optimization methods within ``nabs`` take a similar approach to the
```
ZLLentz: Imagine making an A+ pull request and forgetting the 1-line doc summary as the first line convention

nabs/optimize.py
```diff
@@ -0,0 +1,289 @@
...
+                         'method': method,
+                         'maximize': maximize},
+           'plan_name': 'optimize',
+           'hints': {}}
+    try:
+        dimensions = [(motor.hints['fields'], 'primary')]
+    except (AttributeError, KeyError):
+        pass
+    else:
+        _md['hints'].setdefault('dimensions', dimensions)
```
ZLLentz: Can you explain what this dimensions field is?

nabs/optimize.py
```diff
@@ -0,0 +1,289 @@
...
+    else:
+        final_region = (c, b)
+    logger.debug("Extrema determined to be within %r",
+                 final_region)
+    return final_region
+
+
+# Add the optimization docstring to both minimize and maximize
+maximize.__doc__ += optimize.__doc__
+minimize.__doc__ += optimize.__doc__
```
ZLLentz: I'm not sure if you want to dive into the weeds, but there's also a way to edit the `__signature__` attribute so it doesn't just say `*args` and `**kwargs` when you use `?` in `ipython`. Not high prio but I thought you should know.

nabs/utils.py
```diff
@@ -0,0 +1,62 @@
+from ophyd.signal import DerivedSignal
+from ophyd.sim import SignalRO
+
+
+class InvertedSignal(DerivedSignal):
```
ZLLentz: starting to find more candidates for a `pcdsutils` module

ZLLentz:
As an aside, we didn't seem to activate codecov here.

nabs/optimize.py
teddyrendahl: This informs `BestEffortCallback` what the "independent" variables are.

https://github.com/NSLS-II/bluesky/blob/d9abfdd6ac1d2d9633503d6ec28aa3b267b7e349/bluesky/callbacks/best_effort.py#L95

nabs/utils.py
teddyrendahl: 👍

nabs/optimize.py
teddyrendahl: This is because I add it to the other functions doc-strings and didn't want two 1-line doc summaries. You think this is a good thing to do or should I just bite the bullet and duplicate the docstring?

nabs/optimize.py
teddyrendahl: Might be nice. I'll look into this when I fill out the docs

nabs/optimize.py
ZLLentz: If this function is included in the sphinx docs it should follow the convention, otherwise it doesn't matter. You can always include this `__doc__` outside of the function and append it to all functions that need it, and have just a one line summary here that gets added to.

teddyrendahl: Yeah. I think that sounds like a better option. I'll play around with this.

teddyrendahl:
Setup the docs for this repo. This should be ready for final review. Sweet gif coming in later PR...

PR #17, opened 2018-07-10
ENH: monitor_scan

ZLLentz:
## Description

Add full plan `monitor_scan` and the intermediates/helpers I needed to implement it. 

I'm not entirely in love with this implementation, but it's good enough for discussion. The problem is that the monitored event streams are not correlated with other points, so, for example, a callback will need to pick which motor positions to match the monitored stream with to do a proper plot. I think callbacks in general for monitored values will need to be slightly smarter.

I'd want to pair this with a plotting callback that plots the scattered range up the y-axis against the last read motor position on the x axis.

The documents look something like like:
> start
> descriptor(all motors)
> event(all motors)
> descriptor(det1)
> event(det1)
> descriptor(det2)
> event(det2)
{
> event(det1)
> event(det2)
} x N - 1
> event(all motors)
...
lots of repeating
...
> end

I may also need to do some additional manipulation of the metadata dictionary...

## Motivation and Context

https://github.com/pcdshub/hutch-python/issues/118
The idea is to be able to take data over a time period or to have more efficient scans by monitoring up to an events count instead of polling.

## How Has This Been Tested?

Added tests for coverage

Needs to be written

### DIFF
--- a/nabs/plan_stubs.py
+++ b/nabs/plan_stubs.py
@@ -82,6 +82,43 @@ def move_per_step(detectors, step, pos_cache):
     yield from wait_uid(group=grp)
 
 
+def monitor_events(detectors, events=None, duration=None):
+    """
+    Monitor for value updates from detectors and wait for completion.
+
+    This can run for a fixed number of events, for a fixed duration, or both.
+    If both, this will run until both conditions are met: until enough time has
+    passed and until enough events have been seen.
+
+    If events is used, we'll wait until every detector reaches the set events
+    count.
+
+    Parameters
+    ----------
+    events: ``int``, optional
+        If provided, we'll monitor until we have enough events.
+    duration: ``float``, optional
+        If provided, we'll monitor for this fixed duration in seconds.
+        If both arguments are provided, then we'll wait for both to be true,
+        so we'll have a minimum elapsed time and a minimum number of events.
+    """
+    if events is not None:
+        counter = CallbackCounterFuture(events, detectors)
+        sub_id = yield from subscribe(counter, 'event')
+
+    for det in detectors:
+        yield from monitor(det)
+
+    if duration is not None:
+        yield from sleep(duration)
+    if events is not None:
+        yield from wait_future(counter.future)
+        yield from unsubscribe(sub_id)
+
+    for det in detectors:
+        yield from unmonitor(det)
+
+
 def monitor_step(events=None, duration=None):
     """
     Create bluesky per_step hook for monitoring detectors at every point.
@@ -106,21 +143,6 @@ def monitor_step(events=None, duration=None):
 
     def inner_monitor_step(detectors, step, pos_cache):
         yield from move_per_step(detectors, step, pos_cache)
-
-        if events is not None:
-            counter = CallbackCounterFuture(events)
-            sub_id = yield from subscribe(counter, 'event')
-
-        for det in detectors:
-            yield from monitor(det)
-
-        if duration is not None:
-            yield from sleep(duration)
-        if events is not None:
-            yield from wait_future(counter.future)
-            yield from unsubscribe(sub_id)
-
-        for det in detectors:
-            yield from unmonitor(det)
+        yield from monitor_events(detectors, events=events, duration=duration)
 
     return inner_monitor_step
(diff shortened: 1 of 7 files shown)


nabs/plan_stubs.py
```diff
@@ -60,3 +64,96 @@ def measure_average(detectors, num, delay=None, stream=None):
...
+    count.
+
+    Parameters
+    ----------
+    events: ``int``, optional
+        If provided, we'll monitor until we have enough events.
+    duration: ``float``, optional
+        If provided, we'll monitor for this fixed duration in seconds.
+        If both arguments are provided, then we'll wait for both to be true,
+        so we'll have a minimum elapsed time and a minimum number of events.
```
teddyrendahl: Before I read the docstring I assumed you would `or` the two conditions. Not sure this is wrong just surprised. The bonus of having it `or` is you essentially use the `duration` as a timeout.

nabs/plans.py
```diff
@@ -0,0 +1,44 @@
...
+    num : integer
+        number of points
+    events : ``int``, optional
+        The number of epics updates to monitor for at each point.
+    duration : ``float``, optional
+        The fixed duration in seconds to monitor for at each point.
+    md : dict, optional
+        metadata
+    """
+    # I was also surprised this became a 2-liner
```
teddyrendahl: 😄

tests/test_plan_stubs.py
```diff
@@ -26,3 +28,47 @@ def measure_plan(detectors):
...
+        det.put(0)
+
+    def bg(loop):
+        logger.debug('bg')
+        for det in detectors:
+            value = det.get()
+            det.put(value + 1)
+        logger.debug(value)
+        if value < 10:
+            loop.call_later(0.1, bg, loop)
```
teddyrendahl: There is an `ophyd.sim` device that updates at a set frequency. Using this might simplify your test here.

https://github.com/NSLS-II/ophyd/blob/f4fdae506ac33b1611471eb7f5a406c3b39e1f0c/ophyd/sim.py#L161

nabs/plans.py
```diff
@@ -0,0 +1,44 @@
...
+    per_step = monitor_step(events=events, duration=duration)
+    yield from scan(detectors, *args, num=num, per_step=per_step, md=md)
```
teddyrendahl: Should probably play the `metadata` game a little here and put a proper plan name, though `scan` already does tell you what its `per_step` hook is 🤔

nabs/plan_stubs.py
```diff
@@ -60,3 +64,96 @@ def measure_average(detectors, num, delay=None, stream=None):
     yield from stub_wrapper(count(detectors, num=num, delay=delay))
     # Return the measured average as a dictionary for use in adaptive plans
     return stream.last_event
+
+
+def move_per_step(detectors, step, pos_cache):
```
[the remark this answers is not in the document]
ZLLentz: Oh I totally agree, I had the same thought while writing it. I might pull it out for them...

nabs/plan_stubs.py
ZLLentz: Interesting, using `or` does make this more intuitive. It's `and` simply because that was the default behavior that fell out when I implemented it the first time, and I don't know how to set a timeout on the future wait. Maybe this would be better implemented by adding a `wait_status` message (something I've wanted for a while) so we have a reasonable way to `or` here.

tests/test_plan_stubs.py
ZLLentz: Thank you, this is much better. I'll simplify.

nabs/plans.py
ZLLentz: Yeah, I was conflicted on exactly what to put into the metadata here because I didn't want to just copy/paste edit code from scan... maybe a `metadata_mutator` would be a clean way to do stuff like this.

ZLLentz:
You're correct about why I didn't monitor the motor, and your suggestion is reasonable.

I finished this last night and realized that, while interesting, it doesn't quite go far enough... I want to let this sit for a bit and see if ideas bubble up. It does need to build up into a generalized event-builder plan, the scan should be event-built properly, and I also want to make it so that passing in the `daq` object accidentally works as a naive user would expect.

Maybe it also needs a supporting stream that matches the monitored detectors with the last motor read and builds "full" events.

ZLLentz:

PR #18, opened 2018-07-18
WIP: DataFrame Callback

teddyrendahl:
## Description

Generalizable callback that builds a DataFrame. Allows you to simply plugin in any function that manipulates that DataFrame and emits the result as a secondary event stream.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -13,6 +13,7 @@ requirements:
 
     run:
       - bluesky 
+      - pandas
       - python
 test:
     imports:
--- a/nabs/streams.py
+++ b/nabs/streams.py
@@ -1,9 +1,76 @@
 """Standard data streams for automated routines"""
-import numpy as np
 from bluesky.callbacks.stream import LiveDispatcher
+from pandas import DataFrame
 
 
-class AverageStream(LiveDispatcher):
+class LiveDataFrame(LiveDispatcher):
+    """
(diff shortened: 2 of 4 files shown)


ZLLentz:
I've been thinking about all of this a bit. I wonder if instead of having an event builder callback as in #19 we extend this class to notice that there are monitor streams (e.g. more than one descriptor) and stash them all into the same data frame. Then we can have stock routines that are expecting certain keys to have matched timestamps and other keys to be forward-filled and do a more efficient event-matching in post with our data frame, rather than doing it live. We lose the ability to do a "raw" plot with an event-built stream live, but it's not clear to me that this is something that we even wanted to do.

Maybe we even start anticipating what bld timestamps look like (they have the same set of 120 ns fields in the timestamp every time) and automatically sorting data sets into the "fast" and "slow" catagories... Or another approach might be to take the monitor streams as "fast" data and the read streams as "slow" data, if the event docs are marked in some way.

ZLLentz:
One more thought: it'd be nice to be able to produce these streams without a `RunEngine` instance at all, since that locks you into having an `asyncio` loop as your main thread.

e.g. some standard way to use the `ophyd` subscribe to generate event docs and gather them into a data frame

teddyrendahl:
Closing this as it has been completely untouched for over a year. I also think the approach may be slightly flawed. If our goal is to read information emitted from the `RunEngine` with `pandas` API we should probably just be using `databroker`.

PR #19, opened 2018-07-18
WIP: Event build from monitor streams

ZLLentz:
We'll return to this later. The goal is be able to `monitor` in `bluesky` and reconstruct normal-looking event documents to be emitted. This isn't implemented yet, but I have some prototypes for doing this with ophyd subscriptions as well.


ZLLentz:

Issue #7, opened 2018-04-17
Use status objects from non-set objects


teddyrendahl:
Ophyd objects may have multiple methods that return status objects that you would want to wait for in a scan. Currently this is difficult to do. A standard solution should be created either as a custom message or in the `bluesky` library itself.

### Example
```python
def home_then_move(motor, val):
     yield from wait_from_call(motor.home)  # Wait for status object from home message. Ignore name
      yield from mv(motor, val)   # Standard wait on a move
```

ZLLentz:
This is actually pretty easy to implement, but there's a philosophical argument to made that these should be done in a setup routine if it's not a `set` command.

Issue #6, opened 2018-04-13
walk_to_pixel


teddyrendahl:
## Description
This is really just a call of [fitwalk](https://github.com/slaclab/pswalker/blob/4d2c0a2b267b86f9716efb8a29662adb4f602172/pswalker/plans.py#L383). Specify a detector, motor an `lmfit.Model` that describes the relationship between the motor and the detector. Have our motor walk to target using the model as a guide.

```python
def fitwalk(detectors, motor, models, target,
            naive_step=None, average=120,
            filters=None, drop_missing=True,
            tolerance=10, delay=None, max_steps=10):
    """
    Parameters
    ----------
    detectors : list
        List of detectors to read at each step of walk
    motor : ``ophyd.Object``
        Ophyd object that supports the set and wait. This should have a one to
        one relationship with the independent variable in the model you plan to
        optimize
    models : list
        List of models to evaluate during the walk
    target : float
        Desired end position for the walk
    naive_step : bluesky.plan, optional
        Plan to execute when there is not an accurate enough model available.
        By default this is ``mv(0.01)``
    average : int, optional
        Number of readings to take and average at each event. Models are
        allowed to take a subset of each reading and average, however if the
        two settings will create an average over multiple steps in the walk the
        :attr:`.LiveBuild.average` setting is automatically updated. For
        example, if the walk is told to average over 10 events, your model can
        either average over 10, 5, 2, or 1 shots.
    filters : dict, optional
        Key, callable pairs of event keys and single input functions that
        evaluate to True or False. For more infromation see
        :meth:`.apply_filters`
    drop_missing : bool, optional
        Choice to include events where event keys are missing
    tolerance : float, optional
        Maximum distance from target considered successful
    delay : float, optional
        Mininum time between consecutive readings
    max_steps : int, optional
        Maximum number of steps the scan will attempt before faulting.
        There is a max of 10 by default, but you may disable this by setting
        this option to None. Note that this may cause the walk to run indefinitely.
    """

def walk_to_pixel(detector, motor, target, filters=None,
                  start=None, gradient=None, models=[],
                  target_fields=['centroid_x', 'alpha'],
                  first_step=1., tolerance=20, system=None,
                  average=1, delay=None, max_steps=None,
                  drop_missing=True):
    """
    Step a motor until a specific threshold is reached on the detector

    This function assumes a linear relationship between the position of the
    given motor and the position on the YAG. While at the onset of the plan, we
    don't know anything about the physical setup of the system, we can track
[...]

PR #21, opened 2020-09-29
ENH: add plans from hxrsnd

### LINKED ISSUE #4: field_prepend
Guesses what the name of a field in an event document will be based on the signal name and object. In general, we should use this less and have objects give us `hints`. This is still useful and also included in many of the `HXRSnD` scans.

```python
def field_prepend(field, obj):
    """
    Prepend the name of the Ophyd object to the field name
    Parameters
    ----------
    field : str
        Field name
    obj : object
        Object with :attr:`.name`
    
    Returns
    -------
    target_field : str
```

klauer:
* Untested
* Mainly just a placeholder PR / indicator that reusable plan-related stuff exists in other repositories and may be useful here.

Closes #4 

Need to update to address or invalidate #6, #7 (tagged hxrsnd)

pswalker imports need to be brought in or removed (related: #22)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,29 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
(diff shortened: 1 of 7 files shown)


klauer:
This is old/stale; closing

PR #31, opened 2020-12-16
MNT: add fixed_target_scan plan

cristinasewell:
## Description

Add a new plan for a fixed target scan has been added as well as one with daq support.
This scan is using three motors, motor `x` and `y` for the motion in x and y direction respectively on the `XYGridStage`. The third motor can be any motor that will be scanned.
Besides the three motors, this scan also requires:
-  a **sample_name**, which is a sample saved with targets from a grid.
-  **path** - where the sample exists
- a list of **detectors** - if any
- a list of points for the third motor to go through
- **n_shots** - which is the number of shots we want to take, for every time we move the third motor to a point.

This scan is using the information from the sample provided, to know which targets should be shot next.

In addition to the scan functions, a little function has been added along with a new file `visualizations.py` to give the user a visual idea of how many targets have been shot - looks like this: (note the scaling is not so great, but they would usually have more numbers of columns than in this example + also can be improved later)
true_target_grid

## Motivation and Context

Used in connection with this guy: https://github.com/pcdshub/pcdsdevices/pull/727

## How Has This Been Tested?

Has been tested locally a lot!!

Ex: using dummy motors: (note i am using the `fixed_target_scan` - but to include daq one would use `daq_fixed_target_scan`)
```
from nabs.plans import fixed_target_scan as tscan
from nabs.visualizations import show_shot_targets
from ophyd.sim import motor1, motor2, motor3, det4
ss = [1, 2]
path = '/my/path/to/a/dummy/samples.yml'
plan = tscan('test_sample', [det4], motor1, motor2, motor3, ss, n_shots=3, path=path)
RE(plan)


Transient Scan ID: 1     Time: 2021-01-28 13:06:54
Persistent Unique Scan ID: '50d1c161-967c-4c92-8372-46e309eb5c2e'
New stream: 'primary'
+-----------+------------+------------+------------+------------+------------+
|   seq_num |       time |     motor1 |       det4 |     motor3 |     motor2 |
+-----------+------------+------------+------------+------------+------------+
|         1 | 13:06:54.2 |    -19.826 |      0.000 |          1 |     27.164 |
|         2 | 13:06:54.2 |    -20.078 |      0.000 |          1 |     27.162 |
|         3 | 13:06:54.3 |    -20.330 |      0.000 |          1 |     27.160 |
|         4 | 13:06:54.3 |    -20.581 |      0.000 |          2 |     27.158 |
|         5 | 13:06:54.3 |    -20.577 |      0.000 |          2 |     27.414 |
|         6 | 13:06:54.3 |    -20.325 |      0.000 |          2 |     27.412 |
+-----------+------------+------------+------------+------------+------------+
generator fixed_target_scan ['50d1c161'] (scan num: 1)

```
The operators have tested this too (except the latest changes) - so we could wait for more feedback - or we can proceed with moving forward with this and add corrections/features later.
The last test I have done (for my last changes) was on `xpp` with dummy motors and `daq` simulator.

With all these testing however - there could still be edge cases that I did not think of probably ... 

Docstrings in the code.
Also have provided operators with simple steps/files for how to use this code.

## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Test suite passes locally


cristinasewell:
@klauer and @ZLLentz  is this somewhat close to what I needed to do for this PR?

klauer:
If the goal is to make a placeholder to later add in more functionality to fulfill the original request, it seems OK, I suppose?

cristinasewell:
> If the goal is to make a placeholder to later add in more functionality to fulfill the original request, it seems OK, I suppose?

Ok so I have nothing basically - that means I don't understand yet what needs to be done. 😞 
will come back to this, thanks

klauer:
I think your snaking-with-transform point generation code probably belongs in this PR, as it's similar to bluesky's plan patterns. After that, there'll be more to add here when you have to recenter each sample with the alignment camera (the "extras" in the request).

I don't think you're so far off. And to be fair, it's very possible _I_ still don't understand what needs to be done ;)

nabs/plans.py
```diff
@@ -699,26 +699,43 @@ def fixed_target_scan(detectors, x_motor, xx, y_motor, yy, scan_motor, ss,
...
-                # pp.get_burst(1)
-                yield Msg('open_run')
-                yield from bps.trigger_and_read(detectors + [x_motor,
-                                                y_motor, scan_motor])
-                yield Msg('close_run')
+        for j in range(n1):
+            yield from bp.list_scan([], scan_motor, [ss[j]])
+            x_pos = xx[(j * n_targets):((j + 1) * n_targets)]
+            y_pos = yy[(j * n_targets):((j + 1) * n_targets)]
+            yield from bp.list_scan(detectors, x_motor, x_pos, y_motor, y_pos)
```
cristinasewell, line 717: Assuming I have `detectors = [seq]` here, the list_scan should do a `trigger` message correct? If this is so, then it will send a trigger to the sequencer and that should call the specific sequence (pp.burst sequence)???

Or do I specifically do a `yield from trigger(detectors)` ?

Or do I have to actually have the pulse picker as a parameter to the scan???

What about the pulse_picker being "optional"? In case the users want to use the sequencer but not the pulse picker, do they have to change the sequence and run the scan again???

cristinasewell: Also do I have to have the second function specific with `daq` or can I just add the `@nbpp.daq_during_decorator(record=record, controls=control_devices)` decorator here for the `inner_scan` like @ZryletTC  mentioned in the chat?

Or do we make it separate because we don't always use `daq`?

cristinasewell: @silkenelson I would like some clarifications about the scan if you could help me - this should be something simple but it is taking me days... 

Should I assume in my scan that I will send an `pp` object, so I can do something like this:
```
if use_pp:
     pp.get_burst(1)
     pp.prepare_burst(xxx) 
```
Or should I assume that i will only have a sequencer, and the sequencer will be running the correct sequence (including the one set for pp).

klauer: The `list_scan` takes care of triggering and reading for you.
How do you figure that out?
`list_scan` yields from `scan_nd`:
https://github.com/bluesky/bluesky/blob/6e799a77db42ad83eebbeb0aa6e24bee24fe99d0/bluesky/plans.py#L176-L177
`scan_nd`, at each step, by default calls `one_nd_step`:
https://github.com/bluesky/bluesky/blob/6e799a77db42ad83eebbeb0aa6e24bee24fe99d0/bluesky/plan_stubs.py#L1087-L1113

`one_nd_step` `take_reading` by default is `trigger_and_read`. `trigger_and_read` is:
https://github.com/bluesky/bluesky/blob/6e799a77db42ad83eebbeb0aa6e24bee24fe99d0/bluesky/plan_stubs.py#L869-L912

Which, you can see, yields 'trigger' and 'read' messages as the name implies. 

Without digging into the source above, you can inspect plans to see what bluesky is doing (https://blueskyproject.io/bluesky/simulation.html).  bluesky's documentation is pretty good, so referring back to it is usually worthwhile.

Or if you're feeling particularly hacky, you could always make a device with a `trigger` method and `print('triggered!')` yourself to see what `list_scan` does to your detectors.

It's a bit tough to read through on the first go, but it should get easier as you get more familiar with it.

cristinasewell:
I am going to paste here some additional info that we need for this scan and might tag some people to make sure I understand correctly:
@silkenelson  - please let me know if I am still not getting this correctly


What we eventually need is a scan like this:

`target_scan(sample=current_sample, start_index=last_shot_index+1, scan_motor, ss)`

where the `current_sample` has all the grid positions (kind of like a database)
and we probably need a parameter to indicate **how many targets** we want to shoot.

The idea is that we want to do something like this:
**Example**: we want to take 5 shots, at 1 time delay, (assuming the `scan_motor` is the delay), and then we want to change that certain motor by a value, and take another 5 shots, and keep doing that.

Similar to what I currently have, except I don't have a starting point - the `last_shot_index + 1` to know where to start shooting from.

So I need to save the `last_shot_index` somewhere, for a specific sample – I also need to save the `sample_grid` and to track somehow what targets have been shot so that I am also able to display those targets that have been shot.

It seems like I need to have some additional functions – mostly the ability to save some more information about the grids for samples and the targets that have been shot - as well as access to the `last_shot_index`. 
@klauer @ZLLentz  @ZryletTC  - I would really appreciate some suggestions about where these functions/storing info should go - and maybe how should I go about storing that if you have any great ideas


I am already saving some information about samples, but that information is related to the `xy` mapping for those targets in the sample.... I also have a function that can compute all the points for a given sample. 
I think I need to have an additional file that saves some grids - (all the points), and the `last_shot_index` 
Or I can reuse the current sample file that has the info to compute a grid from, and maybe save the last index that was shot in there too? - I don't know if that would be a good idea though.

silkenelson:
Unfortunately I am out o the loop a little about the first setup of target shooting, but for the sake of argument I will assume we have a function that in the end results in a list of target positions and their respective positions for each of the motors needed to move the sample from target to target.

In my view biased by the Miller experiment implementation, I would expect that "align/prepare/..." function to maybe write a text file (or create a data base entry) where the name/key is the sample name and which could then contain a list of all the target positions and maybe a binary "shot" indicator that starts at 0.

The target_scan should be passed the sample (so that the file or the db can be read) and whatever other motor the scientists would like to scan, which positions that motor should go to and how many targets we would like to shoot at each scan position, so something like:

target_scan(sample=current_sample, nshots=1, scan_motor, ss)

Here, I have moved nshots before the scan motor because ini theory, they might just want to shoot N targets without changing anything outside, but we'd still want to start the DAQ, program & run the sequences & stop everything once we are done. I am open to discussions of wether that should be a different call, but we should share most of the functionality underneath.

target_scan should then at the end in the cleanup step, rewrite the file and change the binary shot indicator for whichever target positions we have shot. The second call on the same sample should then in the start function figure out where the first not-shot position is and fill the shot position list from there.

cristinasewell:
Ok thank you. I currently have a file where I save the samples information, I am not saving the actual grid points there, but just some information to be able to map these points. The idea was that I could calculate any points on the grid given some information - and  this way we can avoid having a file that gets too large (if saving all the points) - but, calculating the points during a scan adds a bit of delay, not too much, but it does.

I can have a separated file with just the sample names and grids, or I can try to calculate the points during scan (this is what i am trying right now) - and update this file for at least the `last_shot_index`. 
Or I can just add all the grid points to the file too.

silkenelson:
I have a minor preference for keeping the sample locations: this would make the target shooting section of this code still work if we had a oddest shaped target ever (if we e.g. sputtered some sample on a wafer). It would also allow us to add another column that could indicate the quality of the sample: for the Miller experiment, they visually inspected the samples would save the positions that had good crystals so that we would not waste any time shooting positions that did not have a good crystal win them. 
Neither of these cases applies here, so I don't have a big problem leaving the current method in place if there is more important tasks remaining to be done.

cristinasewell:
Ok that makes sense, I will save the grid points then - if we have problems with the file getting too large maybe in the future we can save different files for different samples.

nabs/plans.py
```diff
@@ -653,3 +658,131 @@ def daq_a3scan(m1, a1, b1, m2, a2, b2, m3, a3, b3, nsteps):
...
+
+            x_pos = xx[(_last_index + 1):(
+                (_last_index + 1) + n_shots)]
+            y_pos = yy[(_last_index + 1):(
+                (_last_index + 1) + n_shots)]
+            yield from bpp.stub_wrapper(bp.list_scan(detectors, x_motor,
+                                                     x_pos, y_motor, y_pos))
+            _last_index = _last_index + n_shots
+        yield from bps.close_run()
+        update_sample(sample, path, _last_index)
```
cristinasewell, line 735: I am updating the sample file here - assuming that these targets (`x_pos`, `y_pos`) are shot - but I don't know if they have `truly` been shot - is there a better way here to guarantee that the targets have been shot before updating the file to indicate the last target that has been shot?

ZLLentz: Why don't we know if they have truly been shot? I think if we get here, then at least we've run the sequencer at all of these points, right?

cristinasewell: Yes correct, I guess we just don't know if the sequencer actually let the pulse go through and hit the target.... but that's already something else - i probably should not worry about that here

ZLLentz: Oh! I see what you mean. In some cases it might be possible to figure out if the target was hit via live feedback from detector data (it should be a bright hit, not a dark, for example) but maybe this is more complexity than is worthwhile for this.  It's not clear to me exactly what data we'd be having to work with, for example, and it may change from experiment to experiment.

Another thing to think about is interrupted scans; what do we do in that case? Do we mark the first n targets as hit if we got that far?

cristinasewell: Yeah we want to mark them as shot if they were indeed shot.... somehow
Updating the file a little earlier will probably help with this - and maybe updating after each target is scanned - but this will add some delay that we might not want, and maybe other problems....

I need to do a bit of research about what Bluesky offers in this situation, if we know the last point that has been scanned we can find the index of that point, or we can get the motor position and find the index based on that. Here I would need to know when a scan actually got interrupted.

cristinasewell: I added a `try/except` block - and I am able to catch when we call `RunEngine.stop()` and `RunEngine.abort()` - the `RunEngine.resume()` should be fine as is - and the `RunEngine.halt()` - I don't know how to deal with this one, since this guy is not doing any cleanup - if we call this we might have to manually update the last_shot_index?

cristinasewell:
I also added a little function in `nabs` that displays a plot of `shot` targets .... but this is kind of lame, if we have a lot of targets it is hard to see which ones are shot, I don't know how helpful it will be. 
Should I be using something else for this little window display - like `typhos`? 
Also I don't know where such function belongs - I placed it in `nabs.plan_stubs` for now - but it could belong in the `XYGridStage` object?

It displays something like this for now: 
show_grid_plot

nabs/plans.py
```diff
@@ -653,3 +659,138 @@ def daq_a3scan(m1, a1, b1, m2, a2, b2, m3, a3, b3, nsteps):
...
+                         f' bigger than the available samples: {len(xx)}. '
+                         'Please provide a number in range.')
+
+    def inner_scan():
+        global _last_index
+        try:
+            yield from bps.open_run(md={})
+            for i in range(len(ss)):
+                yield from bps.mv(scan_motor, ss[i])
+                yield from bps.read(scan_motor)
```
ZLLentz, line 724: I think this line no longer does anything

nabs/plans.py
```diff
@@ -653,3 +659,138 @@ def daq_a3scan(m1, a1, b1, m2, a2, b2, m3, a3, b3, nsteps):
...
+                    (_last_index + 1) + n_shots)]
+                y_pos = yy[(_last_index + 1):(
+                    (_last_index + 1) + n_shots)]
+                yield from bpp.stub_wrapper(bp.list_scan(detectors, x_motor,
+                                            x_pos, y_motor, y_pos))
+
+                _last_index = _last_index + n_shots
+            yield from bps.close_run()
+            update_sample(sample, path, _last_index)
+        except Exception:
```
ZLLentz, line 736: `bluesky` does have some built-in tools for plans that do "something" after being aborted, see https://github.com/bluesky/bluesky/blob/4d8240e160ab0ba78652848c595e3b431a0c6175/bluesky/preprocessors.py#L469` and as an example https://github.com/bluesky/bluesky/blob/4d8240e160ab0ba78652848c595e3b431a0c6175/bluesky/preprocessors.py#L916

E.g. you definitely want to do `bps.close_run()` regardless of when the scan breaks, usually this would be done for you in https://github.com/bluesky/bluesky/blob/4d8240e160ab0ba78652848c595e3b431a0c6175/bluesky/preprocessors.py#L307

nabs/plan_stubs.py
```diff
@@ -67,3 +69,105 @@ def measure_average(detectors, num, delay=None, stream=None):
...
+        last_shot_index = sample['last_shot_index']
+        xx = sample['xx']
+        yy = sample['yy']
+        return m_points, n_points, last_shot_index, xx, yy
+    except Exception:
+        err_msg = (f'This sample {sample_name} might not exist in the file.')
+        raise Exception(err_msg)
+
+
+def snake_grid_list(points):
```
ZLLentz, line 146: This is good and doesn't need a refactor, but you should be aware that something similar is already built into `bluesky`: https://github.com/bluesky/bluesky/blob/4d8240e160ab0ba78652848c595e3b431a0c6175/bluesky/plan_patterns.py#L302

ZLLentz:
I think the above plot should be evaluated by the end-user, the scientists. I think it looks neat (sizing problems aside for the 4x200 grid)

nabs/plans.py
cristinasewell: Yes that is correct - thank you, I removed it in 7ac3131

nabs/plans.py
cristinasewell: Ohh I have looked at this `finalize_wrapper` a bunch of times before and did not think that it does what it does - that's very nice - thank you for sharing these.

I am using now in 7ac3131 the `finalize_wrapper` and added a callable: `finalize()` to do the updating of the index. 
I am also using the `run_decorator` now, so I took out the extra `open_run`/`close_run` Msgs
.. hopefully I got your suggestions right - please let me know

ZLLentz: I think this looks very clean now

nabs/plan_stubs.py
cristinasewell: Ok tahnk you, I probably won't have time to refactor this now, but I looked into this and it seems pretty straight forward - except that I will have to deal with cyclers - so I will have to probably use `scan_nd` directly instead of list_scan? (unless i want to extract back the lists from the cyclers)

ZLLentz: I haven't dug into the details, I just remembered "something built-in definitely does this" and I thought I'd mention it here.
And yep, no need to refactor at this stage- if it works, it works.

cristinasewell:
> I think the above plot should be evaluated by the end-user, the scientists. I think it looks neat (sizing problems aside for the 4x200 grid)

The scientists will be fine with something like this for now, I checked with Diling - but where should I place this little function? I wrote it in `plan_stubs` - should I keep it here? or should I make it part of `XYGridStage` object?

ZLLentz:
That's a really good question.

If you think it'll only be used in conjunction with the plan, keeping it in this repo is fine- I wonder if we need to open a new file for visualizations?

If you think it may be useful outside of the plan (any time we have this object), then it probably belongs in `pcdsdevices`

nabs/plans.py
cristinasewell: Ohh one thing here that I noticed is that the `summarize_plan` will change the `last_shot_index` as well.... do scientists use this functionality usually?

cristinasewell:
> That's a really good question.
> 
> If you think it'll only be used in conjunction with the plan, keeping it in this repo is fine- I wonder if we need to open a new file for visualizations?
> 
> If you think it may be useful outside of the plan (any time we have this object), then it probably belongs in `pcdsdevices`

I think from an end-users perspective - they would have probably preferred for everything to be part of one object, including the plan... but of course this might not be practical.
I thought of placing it in `nabs` first because `bluesky` has the plots for scans as well - and that seemed appropriate. 
Now I am thinking that it will be more useful if it is in the `XYGridStage` object - there we already have the `path` and we have access to the samples - but I can also import this function in `pcdsdevices` and use it in the object if I need to. 


Here is another problem though that adds up to this one: - i would really like some suggestions here:

I am saving the x and y points in the `sample.yaml` file - and I am not saving them in a `snake_like` pattern - I moved this option in the scan-function. 
When I create the plot for visualizing the shot targets however, I need to know if they scanned it in a `snake_like` pattern to be able to plot it correctly.
Do you think I should have given the option for the user to save it directly into the file as `snake_like` instead? - for now I have this: `def show_shot_targets(sample_name, path, snake_like=True):` where the user has to know if this sample was shot in a `snake_like` pattern or not - i don't think that's very good - but I also don't like the idea of saving them in the file directly as `snake_like`- what if they don't want to scan it this way?

silkenelson:
I think it's ok to expect the user to save the sample in the way that they will intend to shoot it - I don't expect that they will redefine halfway how they'll do this and I don't see how else last_index_shot will really make any sense. Or maybe this is a point where instead of 'last_index_shot' we actually save the shot status for each sample position (where we could then also optionally save other information like 'contains crystals' or whatever we would use this for.

nabs/visualizations.py
```diff
@@ -0,0 +1,67 @@
...
+    ----------
+    sample_name : str
+        The name of the sample file to plot the graph for.
+    path : str
+        The path of the sample file.
+    snake_like : bool
+        Indicates if the targets have been shot in a snake_like pattern.
+        # TODO this is a bad assumption - but most likely they will be
+        shot in a snake-like pattern....
+    """
```
ZLLentz, line 23: This docstring needs some reference to what it's used for- like the specific class it's designed for. Otherwise, coming across this while browsing the code is very confusing.

ZLLentz: Also, I'm starting to see what you meant when you said you thought all of this stuff should be bundled together... hmm

cristinasewell: I updated a bit the docstring for this function (97f39c1) - I will update it further if/when I change the `snake_like` pattern situation.

cristinasewell:
Yes @silkenelson it makes sense to save them the way we want to shoot them - and this is simple to do. And that's correct, if they switch in the middle of the target sample to `non snake_like` pattern - we might run into problems with the `last_shot_index` 
It is also not complicated to add an additional info for each target. 

I am going to proceed with adding that additional info (`0`-available, `1`-shot) - but won't push the changes yet? - Diling and Takahiro were testing the code yesterday night, and they will give me some feedback - as long as they don't switch in the middle with running different patterns, they should be ok - so before I make these changes maybe it is worth having something merged that is working (if Diling and Takahiro say it is working as expected) and then adding the additional info to the targets afterwards?

silkenelson:
This sounds reasonable.

cristinasewell:
Added the `status` for each target - I had to push these changes to be able to test them - if something goes wrong here I think I could just reset to previous commit.
But if this works I think I should stop changing/adding things to these two PRs related to it so we can merge them - and then later we can refactor things.

nabs/plans.py
ZLLentz: I doubt this is used as much as it should be, but it's worth considering how to deal with it

cristinasewell: ohh i agree - I just can't think of how I would do that... 

Please let me know if you have anything in mind.

From my scan I can't know if `summarize_plan` will be called - at least I don't know how I would do that. 

If I could magically transform my `update_sample` method to be like a `Msg.command` - but even then, I don't know how I would tell my function not to do anything when `summarize_plan` is called....

I will do more research here...

cristinasewell: I can do something like this:
```
   import ast
    import inspect
    call_names = [c.func.id for c in ast.walk(ast.parse(inspect.getsource(summarize_plan)))
              if isinstance(c, ast.Call)]
    if 'print_summary_wrapper' in call_names:
        print('FOUND IT..........')
```
and go from there... but I have not used this before and I don't know how save that would be to include in a scan function...

ZLLentz: Yeah I'm also not sure the "best" way to accomplish this. To make the `bluesky` scan work properly in `summarize_plan`, ALL operations that make some change must be done as messages... Your approach seems like it would work to me, but I'm not really sure if we should be inspecting the call stack as part of a scan.

An alternative might be something like, always using `Signal` objects to hold values and write out files, etc... I'm not sure what built-in tooling exists for this. Any thoughts @klauer?

At any rate, this doesn't need to be a priority if it's holding up the project.

nabs/plans.py
```diff
@@ -653,3 +659,130 @@ def daq_a3scan(m1, a1, b1, m2, a2, b2, m3, a3, b3, nsteps):
...
+                         f' bigger than the available samples: {len(xx)}. '
+                         'Please provide a number in range.')
+
+    @bpp.run_decorator()
+    def inner_scan():
+        try:
+            for i in range(len(ss)):
+                yield from bps.mv(scan_motor, ss[i])
+                for j in range(n_shots):
+                    x = (next(item['pos']
```
cristinasewell, line 722: I am a bit concern here that it will add some extra delay because the new way of getting the next targets to be shot is a bit more complex. Could I have done this a bit more efficiently?

nabs/plans.py
```diff
@@ -653,3 +659,130 @@ def daq_a3scan(m1, a1, b1, m2, a2, b2, m3, a3, b3, nsteps):
...
+                    y = (next(item['pos']
+                         for item in cycle_yy if item["status"] is False))
+                    yield from bpp.stub_wrapper(bp.list_scan(detectors,
+                                                x_motor, [x], y_motor, [y]))
+            update_sample(sample, path, (n_shots * len(ss)))
+        except Exception:
+            current_position = x_motor.position
+            try:
+                last_index = next((index for (index, d) in enumerate(xx)
+                                  if np.isclose(d["pos"], current_position)))
```
cristinasewell, line 733: I had to use `np.isclose` to try to find the current position in the saved sample. I am using this in case we stop a run, and based on the current motor position I try to see which target has been shot last. This works in theory if the motor's readback position is exactly the same as saved in the file, but this is not the case all the time - how likely is that `np.isclose` will still not be able to match a saved position in the file?

nabs/plans.py
```diff
@@ -653,3 +659,130 @@ def daq_a3scan(m1, a1, b1, m2, a2, b2, m3, a3, b3, nsteps):
...
+    n_shots : int
+        Indicates how many shots should be taken, or how many samples should
+        be scanned on the grid.
+    path : str
+        Path where the sample file is located.
+    """
+    detectors = list(detectors) + [scan_motor]
+
+    xx, yy = get_sample_targets(sample, path)
+    cycle_xx = cycle(xx)
```
cristinasewell, line 699: I don't have a lot of experience with `cycle` - doing a bit of research I tried to use it to be able in the scan and it works, but any of you are more familiar with it - are there any concerns I should be aware of using `cycle`?

nabs/plan_stubs.py
```diff
@@ -67,3 +68,104 @@ def measure_average(detectors, num, delay=None, stream=None):
...
+    yy = info[1]
+    # find the index of the targets that is next to be shot
+    x_index = next((index for (index, d) in enumerate(xx)
+                    if d["status"] is False), None)
+    if x_index is None:
+        raise IndexError('Could not get a target index that has not been shot,'
+                         ' probably all targets were shot from this sample?')
+
+    temp_x, temp_y = [], []
+    for i in range(n_shots):
```
cristinasewell, line 104: I am trying to update the saved sample information here, and the way I do it is by knowing how many targets have been shot, and rewrite the information for those targets with the new Status - this can be tricky because I need to make sure I don't add additional targets to the file or I don't take any out or I don't override the wrong ones. Testing this seems like it works fine - but at first glance, are there any edge cases or concerns that I did not consider yet?

nabs/visualizations.py
```diff
@@ -0,0 +1,56 @@
+import matplotlib.pyplot as plt
+from .plan_stubs import get_sample_targets
+
+
+def show_shot_targets(sample_name, path):
```
cristinasewell, line 5: Previously I was plotting the targets based on MxN values: 10 x 5 grid for example - and the problem here was that I needed to know if we used snake-like pattern or not. Now, I am plotting the actual xy points and this solves that but it would have been nice to also be able to do the MxN grid - is there a nice way to take those `xy` points in `matplotlib` and convert them to only their MxN index on the grid?
From this: (their x, y positions)
true_target_grid
To this: (their MxN values (M=rows, N=columns))
show_grid_plot


Please note these plots are using different data.

nabs/plans.py
cristinasewell: Also one more concern here is what if I have two targets that have the same position saved - should I also check if it has been shot?

nabs/plans.py
klauer: Looks like I missed the call-out on this one - sorry!

I think the proper approach is as Zach mentioned: such a thing likely should be a custom `Msg`. The summary would only then list the message and arguments.

That said, stashing away information in a Python-only signal is pretty useful, too. It exposes the state in a way that can easily be introspected.

nabs/plan_stubs.py
klauer: Though of no real benefit here, the big advantage to bluesky's approach is that it scales to > 2 motors.

nabs/plans.py
cristinasewell: Ok sounds good. I think I'll go the custom `Msg` route (just because I might be able to find more info on how to do this...)
For this pass I'll just have to make sure I tell the scientists not to use the `summarize_plan` nor the `plot_raster_path`.....

nabs/plan_stubs.py
ZLLentz: I can't really follow this part of the code- but I don't see any red flags

nabs/plans.py
ZLLentz: `cycle` is an infinitely long iterator, so as long as that is what you're going for it should be fine

nabs/plans.py
ZLLentz: I'm not sure what in here specifically would contribute to a slowdown. I suspect the network packets are still slower.

nabs/plans.py
ZLLentz: Custom messages have their own quirks to be aware of- mainly, you need to teach the RunEngine how to interpret them, adding some startup configuration cost.

nabs/plans.py
ZLLentz: `np.isclose` should return true if the values are within an absolute tolerance of 1e-8 and an relative tolerance of 1e-5: https://numpy.org/doc/stable/reference/generated/numpy.isclose.html#numpy.isclose

If the motor has moved at all, you won't be able to match the position. Otherwise it should be okay. You can tune the atol and rtol parameters to work as you'd like if it turns out to be not okay.

I don't understand the question about having two targets with the same position.

nabs/visualizations.py
ZLLentz: I don't think there is a matplotlib built-in for "un-skewing" the lines here- you probably just have to track the columns yourself if you want to build something like this.

ZLLentz:
Ok I'm going to merge everything here and in https://github.com/pcdshub/pcdsdevices/pull/727 since data was taken!

We can revise in follow-ups. Good work on this one!

PR #36, opened 2021-02-10
MNT: Fixed Target Stage - add a couple of general scans

cristinasewell:
## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Test suite passes locally

### DIFF
--- a/nabs/plans.py
+++ b/nabs/plans.py
@@ -731,7 +731,10 @@ def fixed_target_scan(sample, detectors, x_motor, y_motor, scan_motor, ss,
         raise IndexError('Could not get a target index that has not been shot,'
                          ' probably all targets were shot from this sample?')
 
-    if ((next_index - 1) + (n_shots * len(ss))) >= len(xx):
+    # TODO this are changes made by Takahiro, keeping here for now
+    # if ((next_index - 1) + (n_shots * len(ss))) >= len(xx):
+    # if ((next_index - 1) + (n_shots * len(ss))) >= len(xx):#for only Run195
+    if ((next_index - 2) + (n_shots * len(ss))) >= len(xx):
         raise IndexError('The number of last shot + n_shots * len(ss): '
                          f'{next_index - 1} + {n_shots} * {len(ss)}= '
                          f'{next_index - 1 + (n_shots * len(ss))}: '
@@ -807,7 +810,7 @@ def fixed_target_multi_scan(sample, detectors, x_motor, y_motor, scan_motor,
         raise IndexError('Could not get a target index that has not been shot,'
                          ' probably all targets were shot from this sample?')
 
-    if ((next_index - 1) + (len(ss))) >= len(xx):
+    if ((next_index - 1) + (n_shots * len(ss))) >= len(xx):
         raise IndexError('The number of next to be shot + n_shots * len(ss): '
                          f'{next_index} + {len(ss)}= '
                          f'{next_index + (len(ss))}: '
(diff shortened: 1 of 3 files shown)


nabs/plans.py
```diff
@@ -904,3 +907,142 @@ def inner_daq_fixed_target_scan():
...
+                         f' requested: {n_targets}')
+    # try to find the m_end and n_end points based on
+    # how many targets the user wants to shoot - the shooting will
+    # be done row by row
+    temp_x = np.arange(1, m_points + 1, 1)
+    temp_y = np.arange(1, n_points + 1, 1)
+
+    arr = [[i for i in temp_y] for j in temp_x]
+
+    def row_col():
```
cristinasewell, line 963: This logic here was a bit more complicated than I though - but it might be that I just over complicated.... who knows...
Basically, if given a start M and N position, and the number of targets to shoot - for some reason I could not think of anything easy and clean to calculate what should our next m and next n be.... 
It can't just be `m_start+1`, `n_start+1` - because this won't shoot them in a row, and from what I have observed they usually want to shoot in a row pattern.....

Also I have a little bug here with the very last target that I want to shoot - I get a  `RuntimeError: generator raised StopIteration` - and it should be because of an index somewhere... but can't figure it out, will look more tomorrow....

cristinasewell: I also did not add the `snake_like` option yet - because I don't know how it will behave if I want to start in the middle of the row somewhere.....

ZLLentz: Perhaps we can calculate all the points, then start from the first x,y point that matches, then proceed only the number of shots forward?

I agree `snake_like` may be tricky here, but maybe it's OK to just define the pattern as if they started from the top? E.g. just generate the list of full points assuming we've been doing `snake_like` all along?

cristinasewell: I think I fixed the bug that I was having. 
But yes, if we want to do snake_like I'm going to have to generate all points first like you suggested, I think I should check with scientists if they want that - because I remember they were saying that they had to spend some time figuring out which row is going to be reversed so they can start at a different point....

Anyway, I now discovered that I have another issue 😭  - with the `extra_motor_scan` - it is not doing what the other scan is doing:
Original Scan - for example:
```ss = [1, 2]```
let's assume x and y saved positions are the following
```x -> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]```
```y -> [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]```
```
plan = tscan(stage.current_sample, [det4], motor1, motor2, motor3, ss, n_shots=3, path=stage.current_sample_path)

RE(plan)
This plan will go the following :
=================================== Open Run ===================================
motor3 -> 1
motor1 -> 1
motor2 -> 1
  Read ['det', 'motor3', 'motor1', 'motor2']
motor1 -> 2
motor2 -> 2
  Read ['det', 'motor3', 'motor1', 'motor2']
motor1 -> 3
motor2 -> 3
  Read ['det', 'motor3', 'motor1', 'motor2']
motor3 -> 2
motor1 -> 4
motor2 -> 4
  Read ['det', 'motor3', 'motor1', 'motor2']
motor1 -> 5
motor2 -> 5
  Read ['det', 'motor3', 'motor1', 'motor2']
motor1 -> 6
motor2 -> 6
  Read ['det', 'motor3', 'motor1', 'motor2']
================================== Close Run ===================================
```

Current Scan: something like this - repeating the same points targets..... I either have a bug here or have to combine these two scans together
```
extra_points = [1, 2]
plan = list(nbp.extra_motor_scan(dets=[hw.det4], stage=stage,
                                     start_m=1, start_n=1,
                                     n_shots=1, n_targets=2,
                                     extra_motor=hw.motor3,
                                     extra_points=extra_points))

=================================== Open Run ===================================
motor3 -> 1
motor1 -> 1
motor2 -> 1
  Read ['det', 'motor3', 'motor1', 'motor2']
motor1 -> 2
motor2 -> 2
  Read ['det', 'motor3', 'motor1', 'motor2']
motor3 -> 2
motor1 -> 1
motor2 -> 1
  Read ['det', 'motor3', 'motor1', 'motor2']
motor1 -> 2
motor2 -> 2
  Read ['det', 'motor3', 'motor1', 'motor2']
================================== Close Run ===================================
```

cristinasewell: Just checked - we do want to have `snake_like` - and it is ok the way you suggested @ZLLentz  - thank you

cristinasewell:
I think this works fine now, needs some testing on dev, but locally seems to work well now.

cristinasewell:
Just going to paste some feedback here for reference from slack channel `lcls-xpplu8318` 

Improvement sample scan script for more flexibilities
      -     Start from (m, n) + ignore the shot log
      -     Start from -1 (the next spot of the last shot).
      -     API like functions for building own scripts

PR #37, opened 2021-02-11
DOC: add release notes for v1.1.0

ZLLentz:
## Description

Last tag notes for the day, @cristinasewell's scanning code as used at the beginning of the experiment. Next tag will be the as-used-at-the-end version.


cristinasewell (approved):
Thanks @ZLLentz

PR #39, opened 2021-02-18
DOC: Fix docs

### LINKED ISSUE #38: Current Build Docs Missing Plans API/Docstrings
## Expected Behavior

Should include docs from the plans.py, etc.

## Current Behavior

See https://pcdshub.github.io/nabs/v1.1.0

## Possible Solution

Investigate docs source folder and fix as needed

## Context

Cannot show docs to scientists

ZLLentz:
## Description

Small edits and fixes to the documentation
This clears all warnings except a few that I don't think I can get rid of that I shouldn't worry about, plus all the intersphinx warnings from "I want to link to this specific page in the ophyd docs but it doesn't exist yet"

Also adds `ophyd` as a runtime dependency because while combing through the build information I noticed that it is needed to import `nabs.utils`.

The key here was pinning the `bluesky` version, since `nabs` uses a specific function introduced in `v1.6.5`. Intermittent failures can be traced back to the wrong version of `bluesky` being installed.

## Motivation and Context

closes #38 

There are some issues preventing the proper docs build from being uploaded.

## How Has This Been Tested?

This passes once there are no import errors inside of the docs travis build.

This IS docs

## Pre-merge checklist

### DIFF
--- a/nabs/plan_stubs.py
+++ b/nabs/plan_stubs.py
@@ -34,7 +34,7 @@ def measure_average(detectors, num, delay=None, stream=None):
         Time delay between successive readings. See `bluesky.plans.count`
         for more details
 
-    stream : `nabs.streams.AverageStream`, optional
+    stream : :py:class:`nabs.streams.AverageStream`, optional
         If a plan will call `measure_average` multiple times, a single
         ``AverageStream`` instance can be created and then passed in on each
         call. This allows other callbacks to subscribe to the averaged data
@@ -74,15 +74,15 @@ def update_sample(sample_name, path, n_shots):
     """
     Update the current sample information after a run.
 
-    Updates the `status` values of each target in the sample,
-    from `False` to `True` to indicate that it is shot.
+    Updates the ``status`` values of each target in the sample,
+    from ``False`` to ``True`` to indicate that it is shot.
 
     Parameters
     ----------
     sample_name : str
         A name to identify the sample grid, should be snake_case style.
     path : str
-        Path to the `.yml` file. Defaults to the path defined when
+        Path to the ``.yml`` file. Defaults to the path defined when
         creating this object.
     n_shots : int
         Indicates how many targets have been shot.
@@ -131,7 +131,7 @@ def update_sample(sample_name, path, n_shots):
 
 def get_sample_targets(sample_name, path):
     """
-    Get the `xx` and `yy` target information from a saved sample.
+    Get the ``xx`` and ``yy`` target information from a saved sample.
 
     Given a sample name, get the x, y grid points that are mapped for that
     sample.
@@ -140,13 +140,13 @@ def get_sample_targets(sample_name, path):
     ----------
     sample_name : str
         The name of the sample to get the mapped points from. To see the
-        available mapped samples call the `mapped_samples()` method.
+        available mapped samples call the ``mapped_samples`` method.
     path : str, optional
         Path to the samples yaml file.
 
     Returns
     -------
-    `xx`, `yy` : tuple
+    ``xx``, ``yy`` : tuple
         Returns two lists of dictionaries, with information about the targets.
     """
     data = None
--- a/nabs/plan_stubs.py
+++ b/nabs/plan_stubs.py
@@ -8,10 +8,10 @@ are intended as building blocks for other complete plans.
 """
 import logging
 
+import yaml
 from bluesky.plan_stubs import subscribe
 from bluesky.plans import count
 from bluesky.preprocessors import stub_wrapper
-import yaml
 
 from nabs.streams import AverageStream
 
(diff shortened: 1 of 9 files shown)


nabs/plan_stubs.py
```diff
@@ -74,15 +74,15 @@ def update_sample(sample_name, path, n_shots):
     """
     Update the current sample information after a run.
 
-    Updates the `status` values of each target in the sample,
-    from `False` to `True` to indicate that it is shot.
```
klauer, line 78: These should have been OK as-is: 
`Built-in constants and keywords can be emphasized with backticks. An appropriate reference to python docs with be automatically generated (e.g. `True` or `assert`).` (https://pcdshub.github.io/style.html ref because I forget frequently)

nabs/plans.py
```diff
@@ -295,7 +297,7 @@ def f(detectors: Iterable[OphydObj]) -> Generator[Msg]:
     Note
     ----
     The events, duration, record, and use_l3t arguments come from the
-    `nabs.preprocessors.daq_step_scan_decorator`.
+    :py:func:`nabs.preprocessors.daq_step_scan_decorator`.
```
klauer, line 300: Was this failing: `docs.conf.default_role = 'any'` should make `:py:func:` superfluous.

klauer (approved):
I'd believe you if you said these lines I commented on caused issues, but I would like to think they shouldn't be necessary (and would wonder why...)

nabs/plans.py
ZLLentz: this one created an "ambiguity" error where there were somehow two valid targets for `any`

nabs/plan_stubs.py
ZLLentz: Hmm, bizarre, these definitely were making warnings (unless I messed up here?)

nabs/plans.py
klauer: Weird! No clue how that could possibly be ambiguous (@ZryletTC any thoughts?)

klauer:
No reason not to merge given that this works - don't let the above comments hold this up. Would be good to better understand it though.

nabs/plans.py
ZLLentz: Let me find an older travis output, it might be in there

nabs/plan_stubs.py
ZLLentz: https://travis-ci.com/github/pcdshub/nabs/jobs/482058753#L2916-L2917

nabs/plans.py
ZLLentz: Can't find it, must have not pushed that inbetween version

nabs/plan_stubs.py
klauer: Hmm
```
/home/travis/build/pcdshub/nabs/docs/source/api.rst.rst:4: WARNING: autosummary: failed to import nabs.plans
/home/travis/build/pcdshub/nabs/docs/source/api.rst.rst:4: WARNING: autosummary: failed to import nabs.preprocessors
```

nabs/plans.py
ZLLentz: I'll quickly undo this locally and build

ZLLentz:
```

File: nabs/plan_stubs.py (hunk 1 above)
Comment (ZLLentz) [reply nabs/plan_stubs.py:78]: Yes, that's one of the things fixed in this PR

Comment (klauer) [reply nabs/plan_stubs.py:78]: I was just wondering if it led to those messages

Comment (ZLLentz) [reply nabs/plan_stubs.py:78]: Ahh, actually those were related to this fix: https://github.com/pcdshub/nabs/pull/39/commits/88557a54e52aba03c32ff5f26d56f1d792589c00

File: nabs/plans.py (hunk 2 above)
Comment (klauer) [reply nabs/plans.py:300]: Ohh, so because it creates a document and also a python function reference. What's the actual difference in where it links you though? Page vs `#portion-of-page`?

File: nabs/plan_stubs.py (hunk 1 above)
Comment (ZLLentz) [reply nabs/plan_stubs.py:78]: Double-confirmed that those true/false errors pop back up if I undo the double backticks on true/false

File: nabs/plans.py (hunk 2 above)
Comment (ZLLentz) [reply nabs/plans.py:300]: That's a great question, I didn't even check...

File: nabs/plan_stubs.py (hunk 1 above)
Comment (klauer) [reply nabs/plan_stubs.py:78]: OK, looks like this is because we're missing "Python" in intersphinx:

(From pcdsdevices)
```
intersphinx_mapping = {
    'ophyd': ('https://blueskyproject.io/ophyd', None),
    'python': ('https://docs.python.org/3', None),
    'numpy': ('https://docs.scipy.org/doc/numpy', None)
}
```

ZLLentz: ahh, let me try adding that in

ZLLentz: better after adding this in

nabs/plans.py
ZryletTC: Weird... I haven't seen that before :/
Maybe the docs in our other packages weren't producing the `:doc:` role for some reason? I'd like to look into this but not sure when I'll get the time.

PR #42, opened 2021-04-17
HACK: tweak signature of step plans for bs gui

klauer:
End goal is this:

```
In [2]: inspect.signature(nabs.plans.daq_dscan)
Out[2]: <Signature (detectors, motor, start, end, nsteps, *, events=None, duration=None, record=True, use_l3t=False)>
```

I'd bet there's a one/two liner out there that'll do this. But it's Friday and I'm not finding it. Also did not make an effort to clean this up before pushing because it's already 5:30, and I haven't really even eaten lunch yet. 

(Don't merge until we figure out if this is really the best we can do)

cc @cristinasewell

### DIFF
--- a/nabs/preprocessors.py
+++ b/nabs/preprocessors.py
@@ -8,13 +8,13 @@ and yield messages from a new, modified plan, as well as "decorator"
 functions that can be applied to ``bluesky`` plan functions to return new
 plan functions with modifications.
 """
-import inspect
-
 from functools import wraps
 
 import bluesky.plan_stubs as bps
 import bluesky.preprocessors as bpp
 from bluesky.utils import is_movable, make_decorator
+
+from . import utils
 
 
 def _get_daq():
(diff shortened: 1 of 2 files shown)


nabs/preprocessors.py
```diff
@@ -197,16 +199,32 @@ def daq_step_scan_decorator(plan):
     """
 
     @wraps(plan)
-    def inner(*args, **kwargs):
-        events = kwargs.pop('events', None)
-        duration = kwargs.pop('duration', None)
-        record = kwargs.pop('record', True)
-        use_l3t = kwargs.pop('use_l3t', False)
+    def inner(*args, events=None, duration=None, record=True, use_l3t=False):
         return (yield from daq_step_scan_wrapper(plan(*args, **kwargs),
```
cristinasewell, line 203: I'm looking into this a bit and still trying to make it work with the `bluesky-queueserver` (honestly might have issues because i don't actually run daq - will let you know more tomorrow)
For now though it is complaining about not knowing where `**kwargs` came from here :D  - probably don't need it anymore here?

klauer: Cool, feel free to push to this branch if you have a fix. (If not, it looks like we should just retain the old "`**kwargs`" and `kwargs.pop()`" style but add the signature-modifying stuff for now.)

I'm still not entirely convinced this is the _easiest_ way, but at least it does seem to work for fixing up the signature.

cristinasewell: I honestly have tried a few things without success haha - I think your solution here is great! 
And both of these worked for me:

```
def inner(*args, **kwargs):
        events = kwargs.pop('events', None)
        duration = kwargs.pop('duration', None)
        record = kwargs.pop('record', True)
        use_l3t = kwargs.pop('use_l3t', False)

and the rest of your fix here... 
```

or just add the `**kwargs` in your current one:

```
def inner(*args, events=None, duration=None, record=True, use_l3t=False, **kwargs):

and the rest of your fix here...
```

klauer: Cool, thanks for giving this a go. I cleaned it up a bit; let's see if this still works.
As a "utility" in the new commit it now feels _less_ of a hack, kinda-sorta...

klauer:
@ZLLentz thoughts? Did I go off the deep end here?

ZLLentz:
No I think this is wonderful

ZLLentz (approved):

ZLLentz:
@klauer @cristinasewell is this ready to merge/tag?

klauer:
Nothing else I need/want to add; if it works for Cristina I'd say let's merge it.

cristinasewell:
> Nothing else I need/want to add; if it works for Cristina I'd say let's merge it.

Yes it works, I forgot to pull the last changes and i was working with your initial changes, but i just tried the newest changes and work.

PR #53, opened 2022-02-10
ENH: Add ELog posting callback

tangkong:
## Description
Adds a callback that posts bluesky run information to the ELog automatically.  Currently posts last ipython input at run start, and a copy of the run table as html at run stop

## Motivation and Context
[JIRA ticket](https://jira.slac.stanford.edu/browse/LCLSECSD-324)

## How Has This Been Tested?
Interactively, with new test added to test suite
Test posts to [devlgbk](https://pswww.slac.stanford.edu/devlgbk/lgbk/tsdx00117/#eLog) 
pcds-5.2.0, pcds-5.2.1

Screen Shot 2022-02-16 at 2 08 29 PM

Some docstrings, including example usage.

## Pre-merge checklist

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 flake8
+matplotlib
 pytest
 pcdsdaq
 pcdsdevices
-matplotlib
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,3 +2,4 @@ flake8
 pytest
 pcdsdaq
 pcdsdevices
+matplotlib
--- a/docs/source/upcoming_release_notes/53-ELog_Post_Callback.rst
+++ b/docs/source/upcoming_release_notes/53-ELog_Post_Callback.rst
@@ -1,0 +1,24 @@
+53 ELog Post Callback
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- Add callback to RE for posting last input and run table to ELog.
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
+- klauer
+- zllentz
--- a/docs/source/upcoming_release_notes/53-ELog_Post_Callback.rst
+++ b/docs/source/upcoming_release_notes/53-ELog_Post_Callback.rst
@@ -20,5 +20,4 @@ Maintenance
 Contributors
 ------------
 - tangkong
-- klauer
 - zllentz
--- a/nabs/callbacks.py
+++ b/nabs/callbacks.py
@@ -1,0 +1,89 @@
+"""
+Callbacks for subscription to the Bluesky ``RunEngine``
+
+This is the LCLS counterpart to `bluesky.callbacks`
+
+Callbacks in this module expect and operate on the `start`, `event`,
+`stop`, etc documents.
+"""
+
+import logging
+
+from bluesky.callbacks.core import CallbackBase, make_class_safe
+
+logger = logging.getLogger(__name__)
+
+
+@make_class_safe(logger=logger)
+class ELogPoster(CallbackBase):
+    """
+    Callback to post messages to the elog, primarily related to plan setup and
+    results. This callback relies on and must be instantiated after the
+    BestEffortCallback and HutchELog.
+
+    To subscribe:
+
+    .. code-block:: python
+
+        elogc = ELogPoster(bec, elog)
+        elogc_uid = RE.subscribe(elogc)
+
+    To enable posting for a specific run:
+
+    .. code-block:: python
+
+        RE(plan(), post=True)
+
+    Parameters
+    ----------
+    bec : BestEffortCallback
+        Instance of BestEffortCallback subscribed to the current RunEngine
+
+    elog : HutchELog
+        Instance of HutchELog connected to the desired logbook
+    """
+    def __init__(self, bec, elog):
+        self._bec = bec
+        self._elog = elog
+
+        # default to elog setting
+        self._send_post = elog.enable_run_posts
+
+        # TO-DO:
+        # - add granular switches as optional arguments
+        # - add error handling for misspelled keys
+        # - add default lists for start doc keys, elog tags?
+
+    def start(self, doc):
+        """ Post plan information on start document"""
+        if 'post' in doc.keys():
+            # Override default if key exists
+            self._send_post = doc['post']
+
+        if self._send_post:
+            run_info = str({k: doc[k] for k in ['plan_name', 'plan_args']})
+            logger.info(f"Posting run start information to elog, \
+                            experiment {self._elog.logbooks['experiment']}")
+            self._elog.post(run_info, tags=['plan_info', 'RE'])
+
+        super().start(doc)
+
+    def stop(self, doc):
+        """ Post to ELog on plan close (stop document)"""
+
+        super().stop(doc)
+
+        # need to hold onto BEC, instead of the table it holds?...
+        table = self._bec._table._rows
+
+        # Can be None of table isn't generated (ie. bp.count)
+        if table and self._send_post:
+            logger.info(f"Posting run table information to elog, \
+                            experiment {self._elog.logbooks['experiment']}")
+            self._elog.post(
+                '\n'.join(table),
+                tags=['run_table', 'RE']
+            )
+
+        # Clean up
+        self._send_post = self._elog.enable_run_posts
--- a/nabs/callbacks.py
+++ b/nabs/callbacks.py
@@ -46,9 +46,6 @@ class ELogPoster(CallbackBase):
         self._bec = bec
         self._elog = elog
 
-        # default to elog setting
-        self._send_post = elog.enable_run_posts
-
         # TO-DO:
         # - add granular switches as optional arguments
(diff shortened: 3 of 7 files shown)


nabs/callbacks.py
```diff
@@ -0,0 +1,89 @@
...
+
+    elog : HutchELog
+        Instance of HutchELog connected to the desired logbook
+    """
+    def __init__(self, bec, elog):
+        self._bec = bec
+        self._elog = elog
+
+        # default to elog setting
+        self._send_post = elog.enable_run_posts
```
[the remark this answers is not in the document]
klauer: The default setting isn't actually used, right? It rather gets reset in `start` based on the md kwargs + elog setting.

nabs/callbacks.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # TO-DO:
+        # - add granular switches as optional arguments
+        # - add error handling for misspelled keys
+        # - add default lists for start doc keys, elog tags?
+
+    def start(self, doc):
+        """ Post plan information on start document"""
+        if 'post' in doc.keys():
```
klauer, line 59:
```suggestion
        if 'post' in doc:
```

Or more concisely:

```python
        self._send_post = doc.get("post", self._elog.enable_run_posts)
```

nabs/callbacks.py
```diff
@@ -0,0 +1,86 @@
...
+            # Override default if key exists
+            self._send_post = doc['post']
+        else:
+            self._send_post = self._elog.enable_run_posts
+
+        if self._send_post:
+            run_info = str({k: doc[k] for k in ['plan_name', 'plan_args']})
```
klauer, line 66: Should this be a `json.dumps`? What is the elog post expecting?

klauer (commented):
I think this is looking good! 

I'm also thinking we could also make something that sends a message to the centralized logger. We currently only log exceptions, but this could be a good spot to start doing more.

nabs/callbacks.py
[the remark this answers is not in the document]
tangkong: You're right, this is vestigial from the way I was handling defaults earlier.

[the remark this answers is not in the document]
tangkong: I guess if we get through a plan without hitting a start document, we have bigger problems

nabs/callbacks.py
tangkong: I'll check with Murali, the only supported formats I saw were plain text and html.  we we can submit any arbitrary attachment, but I think we want this information to be easily readable

tangkong: There's a point to be made here that posts to the elog should be human-readable, so a json is maybe not a desirable format.  

There's the option to post to html, maybe we try to auto-format some of that.  Though the run-start data can become rather deeply nested.

dev-requirements.txt
```diff
@@ -2,3 +2,4 @@ flake8
 pytest
 pcdsdaq
 pcdsdevices
+matplotlib
```
ZLLentz, line 5: I feel so petty suggesting this, but the file was originally in alphabetical order
```suggestion
matplotlib
pytest
pcdsdaq
pcdsdevices
```

nabs/tests/test_callbacks.py
```diff
@@ -0,0 +1,60 @@
...
+|         2 | 12:50:56.8 |      0.111 |      0.994 |
+|         3 | 12:50:56.9 |      0.222 |      0.976 |
+|         4 | 12:50:56.9 |      0.333 |      0.946 |
+|         5 | 12:50:56.9 |      0.444 |      0.906 |
+|         6 | 12:50:57.0 |      0.556 |      0.857 |
+|         7 | 12:50:57.0 |      0.667 |      0.801 |
+|         8 | 12:50:57.0 |      0.778 |      0.739 |
+|         9 | 12:50:57.1 |      0.889 |      0.674 |
+|        10 | 12:50:57.1 |      1.000 |      0.607 |
++-----------+------------+------------+------------+"""
```
ZLLentz, line 48: This sort of test is a bit hazardous in that it will break if the LiveTable format changes

ZLLentz (commented):
Added some thoughts/suggestions
Overall I think this is a clean approach

dev-requirements.txt
tangkong: Of all the things flake8 doesn't clean up for me 😢

nabs/callbacks.py
```diff
@@ -0,0 +1,86 @@
...
+            logger.info("Posting run start information to elog")
+            self._elog.post(run_info, tags=['plan_info', 'RE'])
```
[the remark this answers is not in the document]
tangkong: I think exactly what we send to elog here is up for discussion.   Json isn't exactly human-friendly, and the arguments may not immediately inform the reader as to what the plan was.  I'm not entirely sure how to access the raw IPython input, but that'd certainly be useful.  It might be a hook we add to the ipython session rather than the RunEngine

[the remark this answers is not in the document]
ZLLentz: Yeah, it's not actually clear that the behavior I'm doing belongs in nabs. If it did, it would involve an `ipython` argument to the class where we'd send the `ipython` instance and do something similar to https://github.com/pcdshub/hutch-python/blob/master/hutch_python/ipython_log.py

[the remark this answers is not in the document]
klauer: I think adding the last input is a great idea for this. I think hutch-python should offer a quick way of getting that last input string easily since it has to reformat it to look like the prompt: https://github.com/pcdshub/hutch-python/blob/7f55e3db1f7864d16fd6007bda7082c2511df025/hutch_python/ipython_log.py#L115

nabs/tests/test_callbacks.py
tangkong: Hazardous enough to change the test?  I thought this test pattern was useful, though fragile.  It's how bluesky tests their LiveTable callback, interestingly enough

tangkong:
Some significant changes here, with much inspiration taken from bluesky's LiveTable callback with regards to formatting.  Failures shouldn't halt operation, but I think testing with real beamline devices would be nice.

docs/source/upcoming_release_notes/53-ELog_Post_Callback.rst
```diff
@@ -0,0 +1,24 @@
...
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
+- klauer
```
klauer, line 23:
```suggestion
```

I didn't do any of the hard work - no need to credit me, but thanks!

nabs/callbacks.py
```diff
@@ -0,0 +1,166 @@
...
+
+        super().start(doc)
+
+    def descriptor(self, doc):
+        """ Initialize table information """
+        # these always exist
+        self._data = dict(seq_num=[], time=[])
+        for k in doc['data_keys']:
```
klauer, line 115:
```suggestion
        for k, dk_entry in doc['data_keys'].items():
```

nabs/callbacks.py
```diff
@@ -0,0 +1,166 @@
...
+            dk_entry = doc['data_keys'][k]
+
+            if dk_entry['dtype'] not in self._FMTLOOKUP:
+                logger.warn(f'skipping {k}, format not recognized')
+                continue
+
+            self._data.update({k: []})
```
klauer, line 122: `update` is a bit awkward for a single key, would suggest:
```suggestion
            self._data[k] = []
```

nabs/callbacks.py
```diff
@@ -0,0 +1,166 @@
...
+            )
+
+        # this might break for unfilled data, but do we deal with that?
+        for k in self._data_keys:
+            try:
+                dtype = self._format_info[k]
+                style = self._FMTLOOKUP[dtype]
+                self._data[k].append(style.format(doc['data'][k]))
+            except Exception as ex:
+                print(ex)
```
klauer, line 145: Debug print to remove before merging
```suggestion
```

nabs/callbacks.py
```diff
@@ -0,0 +1,166 @@
...
+                logger.warning(f'key: {k} not in event document')
```
klauer, line 146: Could be a different error, like `style` failing to format too

Also, should `self._data[k]` get a value in this scenario, just in case? Or will remaining data be off-by-one?

klauer (commented):
It's hard to argue with the end-result - this looks great in the elog.

I'd recommend some minor style fixes, but all are optional/functionally equivalent. Keep in mind that when we review things, we tend to be fairly critical of even small things. That's not a reflection of your abilities or us gatekeeping, but rather just trying to have more maintainable code and improve everyone's skills.

I would mark this as "Approve" but you have a good number of TODOs in there. Are you going to tackle them now or in a follow-up PR?

nabs/callbacks.py
tangkong: Could you clarify what you mean by "get a value"?  Do you mean be supplied a default value to keep the table aligned?

klauer: Yes, keeping the table aligned with some default value

klauer (approved):

ZLLentz (approved):
This looks good to me too

PR #55, opened 2022-02-28
ENH: Elog poster for ophyd objects and lists thereof

tangkong:
## Description
Added a function (and helper function) for gathering up ophyd objects and posting them to the elog.  
Recursively walks through an iterable of devices and formats their `.status()` output into collapsible blocks.  

## Motivation and Context
[Jira ticket](https://jira.slac.stanford.edu/browse/LCLSPC-124)

Hopefully this will provide another way for scientists to log their beamline state.  

With the new `biological_parents` attribute of `GroupDevices`, we can post entire namespaces without cluttering the post with child devices

## How Has This Been Tested?
Interactively, sample posts using devices in xcs hutch python, a few tests have been added to `test_utils.py`
pcds-5.3.0

Docstrings, which have some example code

image

## Pre-merge checklist

### DIFF
--- a/nabs/_html.py
+++ b/nabs/_html.py
@@ -1,0 +1,108 @@
+"""
+html templates for use in elog post formatting
+"""
+
+collapse_list_head = """<html>
+<head>
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<style>
+.collapsible {
+  background-color: #777;
+  color: white;
+  padding: 10px;
+  width: 100%;
+  border: none;
+  text-align: left;
+  outline: none;
+  font-size: 15px;
+}
+
+.collapsible:hover {
+  background-color: #555;
+}
+
+.collapsible:after {
+  content: '+';
+  color: white;
+  font-weight: bold;
+  float: right;
+  margin-left: 5px;
+}
+
+.active:after {
+  content: "-";
+}
+
+.content {
+  min-width: 100px;
+  overflow-x: auto;
+}
+
+.parent,
+.child {
+  display: none;
+  margin-left: 20px;
+}
+
+.parent.show,
+.child.show {
+  display: block;
+}
+</style>
+</head>
+<body>
+"""
+
+collapse_list_tail = """<script>
+// manage +/- toggle
+var coll = document.getElementsByClassName("collapsible");
+var i;
+
+for (i = 0; i < coll.length; i++) {
+  coll[i].addEventListener("click", function() {
+    this.classList.toggle("active");
+    var content = this.nextElementSibling;
+  });
+}
+
+// manage show/hide behavior
+var allCollapsibles = document.querySelectorAll('.collapsible');
+allCollapsibles.forEach( item => {
+  item.addEventListener("click", function() {
+     if(this.nextElementSibling.classList.contains('show')) {
+       this.nextElementSibling.classList.remove('show')
+     } else {
+       this.nextElementSibling.classList.add('show')
+     }
+  });
+});
+</script>
+
+</body>
+</html>"""
+
+run_table_head = """<!DOCTYPE html>
+<html>
+<head>
+<style>
+table {
+  font-family: arial, sans-serif;
+  border-collapse: collapse;
+  width: 100%;
+}
+
+td, th {
+  border: 1px solid #dddddd;
+  text-align: left;
+  padding: 8px;
+}
+
+tr:nth-child(even) {
+  background-color: #dddddd;
+}
+</style>
+</head>
+<body>
+"""
+
+run_table_tail = """</body> </html>"""
--- a/nabs/callbacks.py
+++ b/nabs/callbacks.py
@@ -13,6 +13,8 @@ from datetime import datetime
 import pandas as pd
 from bluesky.callbacks.core import CallbackBase, make_class_safe
 
+from ._html import run_table_head, run_table_tail
+
 logger = logging.getLogger(__name__)
 
 
@@ -27,7 +29,7 @@ class ELogPoster(CallbackBase):
 
     .. code-block:: python
 
-        elogc = ELogPoster(bec, elog)
+        elogc = ELogPoster(elog, IPython.get_ipython())
         elogc_uid = RE.subscribe(elogc)
 
     To enable posting for a specific run:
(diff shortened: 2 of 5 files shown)


ZLLentz:
I like how this looks

tangkong:
pretty sure this failure is due to travis not picking up the most recent `pcdsdevices` GroupDevices PR.

nabs/_html.py
```diff
@@ -0,0 +1,108 @@
...
+     }
+  });
+});
+</script>
+
+</body>
+</html>"""
+
+run_table_head = """<!DOCTYPE html>
+<html>
```
klauer, line 85: Preface - I'm no html expert by any stretch.

But there's something a bit weird about embedding `<html>` tags _inside_ of an existing html document. (*)
When you view the source of the elog page with these embedded, how is it structured? Do you see `<html> .... <html> ophyd obj info </html> ... </html>`

If each post is embedded in an `iframe` that's a different story (but I'd guess it's not the case).

I think the "proper" solution would be either getting Murali to modify elog pages to add our css classes for these tables, or on a per-element basis set `style=` (which is a bit messy for sure).

I think this probably shouldn't hold up this PR, but we should look into the right way of doing things to ensure future compatibility with browsers viewing these log entries.

(*) https://www.w3schools.com/html/html_css.asp these are the three "acceptable" ways of embedding style

nabs/callbacks.py
```diff
@@ -49,31 +51,9 @@ class ELogPoster(CallbackBase):
...
-}
-
-tr:nth-child(even) {
-  background-color: #dddddd;
-}
-</style>
-</head>
-<body>
-"""
+    _html_head = run_table_head
```
klauer, line 54: Side thought: we might also consider adding jinja templates in the mix one day to make formatting these things a bit easier.

nabs/tests/test_utils.py
```diff
@@ -0,0 +1,48 @@
+from ophyd.device import Component as Cpt
+from ophyd.device import Device
+from pcdsdevices.device import GroupDevice
+
+from nabs.utils import post_ophyds_to_elog
+
+
+class StatusDevice(Device):
+    """ simulate a device with a status method """
+    def status(self):
```
klauer, line 10: For reference: https://github.com/pcdshub/pcdsdevices/blob/aa724f54473fff8873c29bb40d6634ff3157390a/pcdsdevices/interface.py#L265-L270

nabs/utils.py
```diff
@@ -110,3 +116,108 @@ def add_named_kwargs_to_signature(
     )
 
     return sig.replace(parameters=start_params + wrapper_params + end_params)
+
+
+def format_ophyds_to_html(obj, allow_child=False):
```
klauer, line 121: allow_child isn't documented below.

API thought - you might consider just accepting `*objs` instead of sequence-or-instance.

nabs/utils.py
```diff
@@ -110,3 +116,108 @@ def add_named_kwargs_to_signature(
...
+            "<button class='collapsible'>" +
+            f"{parent_name}" +  # should be a namespace name
+            "</button><div class='parent'>" +
+            content +
+            "</div>"
+        )
+        return out
+
+    # check if parent level ophyd object
+    elif (hasattr(obj, 'status') and
```
klauer, line 165: `callable(getattr(obj, "status", None))` perhaps?

nabs/utils.py
```diff
@@ -110,3 +116,108 @@ def add_named_kwargs_to_signature(
...
+
+        for o in obj:
+            content += format_ophyds_to_html(o, allow_child=allow_child)
+
+        # Don't return wrapping if there's no content
+        if content == "":
+            return content
+
+        # HelpfulNamespaces tend to lack names, maybe they won't some day
+        parent_name = getattr(obj, '__name__', 'expand me')
```
klauer, line 152: Maybe something like `"Ophyd status: " + " ".join(o.name for o in obj)` (and truncate beyond 60 chars or something) for the 'expand me' default?

klauer (approved):
The end result speaks for itself here - looks great! 

 have some open technical questions/suggestions, but I think they can be resolved another day.

nabs/_html.py
tangkong: Good points, it seems the elog strips my html and body tags, so I guess there's some parsing going on at the elog end.  

I tried to keep the css/js as minimal as possible, but if we start to post more things with more involved css I can see asking Murali to hold on to our css classes.

nabs/utils.py
tangkong: whoops thanks

nabs/utils.py
```diff
@@ -110,3 +116,119 @@ def add_named_kwargs_to_signature(
...
+            logger.info(f'skipped {str(obj)}, due to Exception: {ex}')
+
+        return content
+
+    # fallback base case (if ignoring obj)
+    else:
+        return ""
+
+
+def post_ophyds_to_elog(elog, objs, allow_child=False):
```
ZLLentz, line 192: Minor nitpick: no type annotations anywhere, though I don't think these are required.

ZLLentz (commented):
All the functionality here is A+, I like it a lot.

The interface is a bit unwieldy from the user perspective, there are a lot of arguments to pass for a convenience interface. I feel like this will probably be wrapped on a per-experiment basis, e.g. they'll have a wrapped `x.post_elog_status()` method with some pre-selected useful devices.

Does this belong in `nabs`? I think if it was in `pcdsutils` or `pcdsdevices` we could include it on our device interface too, e.g. we could implement `im2k0.post_elog_status()`.

tangkong:
> The interface is a bit unwieldy from the user perspective, there are a lot of arguments to pass for a convenience interface. I feel like this will probably be wrapped on a per-experiment basis, e.g. they'll have a wrapped `x.post_elog_status()` method with some pre-selected useful devices.

I'm all for simplifying the interface, but I'm not sure how to find the elog other than `from xxx.db import elog`.  I guess registering and retrieving the elog is something to look into now.

> Does this belong in `nabs`? I think if it was in `pcdsutils` or `pcdsdevices` we could include it on our device interface too, e.g. we could implement `im2k0.post_elog_status()`.

I like this idea, I'm looking into it now.

nabs/utils.py
klauer: We are generally moving toward annotating most new code, @tangkong, just so you know.
It'll take some getting used to, but we're happy to help when you need it.

klauer:
> Does this belong in `nabs`? I think if it was in `pcdsutils` or `pcdsdevices` we could include it on our device interface too, e.g. we could implement `im2k0.post_elog_status()`.

That's a neat idea! As to where: nabs does have a pcdsdevices dependency and not the other way around (mainly confirming that for myself). pcdsutils seems an okay spot for it.

ZLLentz:
I think it's ok if the `elog` is created and registered some time in the `hutch-python` sphere, so long as we've identified a method for registering it.

The raw methods here as laid out by you don't need to be simplified, they span the full set of things we want to do here, but I think there are numerous ways we can wrap them up nicely for the user.

tangkong:
Moving this to pcdsdevices

PR #56, opened 2022-03-16
ENH: Decorator to interpret floats as step sizes and integers as number of steps

tangkong:
## Description
Adds a decorator that modifies the arguments of the 1D scan plans to interpret floats as step size and integers as the number of steps.

## Motivation and Context
[Jira ticket](https://jira.slac.stanford.edu/browse/LCLSECSD-733)

I assumed that:
- `scan(-1, 1, 1.0)` should scan [-1, 0, 1], (end point included)
- `scan(0, 1, 0.3)` should scan [0, 0.3, 0.6, 0.9] (not including end point)

## How Has This Been Tested?
Interactively, some test cases added.  

Docstrings

## Pre-merge checklist

### DIFF
--- a/nabs/plans.py
+++ b/nabs/plans.py
@@ -454,6 +454,7 @@ def daq_list_scan(*args, per_step=None, md=None):
 
 @bpp.reset_positions_decorator()
 @nbpp.daq_step_scan_decorator
+@nbpp.step_size_decorator
 def daq_ascan(detectors, motor, start, end, nsteps):
     """
     One-dimensional daq scan with absolute positions.
@@ -507,6 +508,7 @@ def daq_ascan(detectors, motor, start, end, nsteps):
 @bpp.reset_positions_decorator()
 @bpp.relative_set_decorator()
 @nbpp.daq_step_scan_decorator
+@nbpp.step_size_decorator
 def daq_dscan(detectors, motor, start, end, nsteps):
     """
     One-dimensional daq scan with relative (delta) positions.
--- a/nabs/plans.py
+++ b/nabs/plans.py
@@ -455,7 +455,7 @@ def daq_list_scan(*args, per_step=None, md=None):
 @bpp.reset_positions_decorator()
 @nbpp.daq_step_scan_decorator
 @nbpp.step_size_decorator
-def daq_ascan(detectors, motor, start, end, nsteps):
+def daq_ascan(detectors, motor, start, end, n):
     """
     One-dimensional daq scan with absolute positions.
 
@@ -477,8 +477,9 @@ def daq_ascan(detectors, motor, start, end, nsteps):
     end : int or float
         The last point in the scan.
 
-    nsteps : int
-        The number of points in the scan.
+    n : int or float
+        if int, the number of points in the scan.
+        if float, step size
 
     events : int, optional
         Number of events to take at each step. If omitted, uses the
@@ -502,14 +503,14 @@ def daq_ascan(detectors, motor, start, end, nsteps):
     :py:func:`nabs.preprocessors.daq_step_scan_decorator`.
     """
 
-    yield from bp.scan(detectors, motor, start, end, nsteps)
+    yield from bp.scan(detectors, motor, start, end, n)
 
 
 @bpp.reset_positions_decorator()
 @bpp.relative_set_decorator()
 @nbpp.daq_step_scan_decorator
 @nbpp.step_size_decorator
-def daq_dscan(detectors, motor, start, end, nsteps):
+def daq_dscan(detectors, motor, start, end, n):
     """
     One-dimensional daq scan with relative (delta) positions.
 
@@ -531,8 +532,9 @@ def daq_dscan(detectors, motor, start, end, nsteps):
     end : int or float
         The last point in the scan, relative to the current position.
 
-    nsteps : int
-        The number of points in the scan.
+    n : int or float
+        if int, the number of points in the scan.
+        if float, step size
 
     events : int, optional
         Number of events to take at each step. If omitted, uses the
@@ -556,7 +558,7 @@ def daq_dscan(detectors, motor, start, end, nsteps):
     :py:func:`nabs.preprocessors.daq_step_scan_decorator`.
     """
 
-    yield from bp.scan(detectors, motor, start, end, nsteps)
+    yield from bp.scan(detectors, motor, start, end, n)
 
 
 @bpp.reset_positions_decorator()
--- a/nabs/preprocessors.py
+++ b/nabs/preprocessors.py
@@ -345,6 +345,8 @@ def step_size_decorator(plan):
         elif type(n) is float:
             # interpret as step size
             start, stop = args[2], args[3]
+            # correct step size sign
+            n = np.sign(stop - start) * np.abs(n)
             if np.abs(n) > np.abs(stop-start):
                 raise ValueError(f"Step size provided {n} greater "
                                  "than the range provided "
--- a/nabs/preprocessors.py
+++ b/nabs/preprocessors.py
@@ -341,23 +341,19 @@ def step_size_decorator(plan):
 
         if type(n) is int:
             # interpret as number of steps (default)
-            yield from plan(*args[:4], n, **kwargs)
-
+            yield from plan(*args, **kwargs)
         elif type(n) is float:
             # interpret as step size
             mmin, mmax = args[2], args[3]
-
             if n > (mmax-mmin):
                 raise ValueError(f"Step size provided {n} greater "
                                  "than the range provided "
                                  f"{mmax-mmin}")
-
             step_list = list(np.arange(mmin, mmax, n))
             # new endpoint needed, for cases where
             # (range % step_size) != 0 or to include endpoint
             if np.isclose(step_list[-1] + n, mmax):
                 step_list.append(step_list[-1] + n)
-
             n_steps = len(step_list)
 
             yield from plan(*args[:3], step_list[-1], n_steps,
--- a/nabs/preprocessors.py
+++ b/nabs/preprocessors.py
@@ -332,10 +332,10 @@ def step_size_decorator(plan):
             # do not support num kwarg
             n = kwargs.pop('num')
         else:
-            # assumes (det_list, motor, mmin, mmax, num)
+            # assumes (det_list, motor, start, stop, num)
             n = args[-1]
 
-        if type(n) not in [int, float]:
+        if not isinstance(n, (int, float)):
             raise TypeError("Step size / number of steps is "
                             "neither float nor integer")
 
@@ -344,15 +344,19 @@ def step_size_decorator(plan):
             yield from plan(*args, **kwargs)
         elif type(n) is float:
             # interpret as step size
-            mmin, mmax = args[2], args[3]
-            if n > (mmax-mmin):
+            start, stop = args[2], args[3]
+            if np.abs(n) > np.abs(stop-start):
                 raise ValueError(f"Step size provided {n} greater "
                                  "than the range provided "
-                                 f"{mmax-mmin}")
-            step_list = list(np.arange(mmin, mmax, n))
+                                 f"{np.abs(stop-start)}")
+            step_list = list(np.arange(start, stop, n))
+            if len(step_list) == 0:
+                raise ValueError("Number of steps is 0 with the "
+                                 "provided range and step size.")
+
             # new endpoint needed, for cases where
             # (range % step_size) != 0 or to include endpoint
-            if np.isclose(step_list[-1] + n, mmax):
+            if np.isclose(step_list[-1] + n, stop):
                 step_list.append(step_list[-1] + n)
             n_steps = len(step_list)
 
--- a/nabs/preprocessors.py
+++ b/nabs/preprocessors.py
@@ -304,8 +304,8 @@ def daq_during_wrapper(plan, record=True, use_l3t=False, controls=None):
 
 def step_size_decorator(plan):
     """
-    Grab the last argument (normally step number), and intepret as
-    - step size if float or < 1
+    Grab the last argument (number of steps), and intepret as
+    - step size if float
     - number of steps if integer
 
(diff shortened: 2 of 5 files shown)


tangkong:
I'd appreciate any feedback on the assumptions I made about the way this handles end points and ranges.  I imagine the behavior here is what the scientists would want, but I could be wrong.

nabs/plans.py
```diff
@@ -476,8 +477,9 @@ def daq_ascan(detectors, motor, start, end, nsteps):
     end : int or float
         The last point in the scan.
 
-    nsteps : int
-        The number of points in the scan.
+    n : int or float
```
ZLLentz, line 480: This is clearly a better name than `nsteps` now that it isn't a number of steps, as much as I dislike single-variable names. I can't think of a better alternative.

nabs/plans.py
```diff
@@ -501,13 +503,14 @@ def daq_ascan(detectors, motor, start, end, nsteps):
...
 
-    yield from bp.scan(detectors, motor, start, end, nsteps)
+    yield from bp.scan(detectors, motor, start, end, n)
 
 
 @bpp.reset_positions_decorator()
 @bpp.relative_set_decorator()
 @nbpp.daq_step_scan_decorator
-def daq_dscan(detectors, motor, start, end, nsteps):
+@nbpp.step_size_decorator
```
ZLLentz, line 512: I like that you boiled this down to a decorator :+1:

nabs/preprocessors.py
```diff
@@ -301,4 +302,64 @@ def daq_during_plan():
...
+            # interpret as number of steps (default)
+            yield from plan(*args, **kwargs)
+        elif type(n) is float:
+            # interpret as step size
+            mmin, mmax = args[2], args[3]
+            if n > (mmax-mmin):
+                raise ValueError(f"Step size provided {n} greater "
+                                 "than the range provided "
+                                 f"{mmax-mmin}")
+            step_list = list(np.arange(mmin, mmax, n))
```
ZLLentz, line 352: reviewer note: numpy docs on `arange`
```
(https://numpy.org/doc/stable/reference/generated/numpy.arange.html#numpy.arange)numpy.arange([start, ]stop, [step, ]dtype=None, *, like=None)
Return evenly spaced values within a given interval.

Values are generated within the half-open interval [start, stop) (in other words, the interval including start but excluding stop).
```
some examples to help me review:
```
In [2]: np.arange(0, 1, 0.1)
Out[2]: array([0. , 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9])

In [3]: np.arange(0, 0.35, 0.1)
Out[3]: array([0. , 0.1, 0.2, 0.3])

In [4]: np.arange(0, 0.35, 1.2)
Out[4]: array([0.])

In [5]: np.arange(0.35, 0, 0.1)
Out[5]: array([], dtype=float64)
```

nabs/preprocessors.py
```diff
@@ -301,4 +302,64 @@ def daq_during_plan():
...
+            # new endpoint needed, for cases where
+            # (range % step_size) != 0 or to include endpoint
+            if np.isclose(step_list[-1] + n, mmax):
+                step_list.append(step_list[-1] + n)
+            n_steps = len(step_list)
```
ZLLentz, line 357: Might be a good idea to raise an exception on `n_steps == 0` to catch any other weird shaping here

nabs/tests/test_plans.py
```diff
@@ -388,3 +388,55 @@ def test_daq_fixed_target_multi_scan(RE, daq, hw, sample_file):
     assert len(reads) == 24
     RE(msgs)
     summarize_plan(msgs)
+
+
+def orange(mmin, mmax, num):
```
ZLLentz, line 393: 🍎

nabs/preprocessors.py
```diff
@@ -301,4 +302,64 @@ def daq_during_plan():
...
+        if type(n) not in [int, float]:
+            raise TypeError("Step size / number of steps is "
+                            "neither float nor integer")
+
+        if type(n) is int:
+            # interpret as number of steps (default)
+            yield from plan(*args, **kwargs)
+        elif type(n) is float:
+            # interpret as step size
+            mmin, mmax = args[2], args[3]
```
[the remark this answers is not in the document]
tangkong: Great point.  Do we think we should correct the sign of the step size based on the range provided?  Rephrasing slightly, should:
- `scan(-1, 1, -0.5)`
- `scan(-1, 1, 0.5)`
do the same thing?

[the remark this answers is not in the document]
ZLLentz: the original args semantically are "start", "stop", "num points", so if we maintain the "start", "stop" meaning then the sign of the step size should be ignored entirely in favor of moving "from start to stop in step sizes of n"

nabs/plans.py
klauer: Is `step` (not `nsteps`) ambiguous enough that it could refer to either the position delta or the number of steps?

Just a thought, but I'm inclined to just say to leave it as-is.

nabs/plans.py
klauer: 💯 % - I would not have thought this could be done so cleanly.

nabs/preprocessors.py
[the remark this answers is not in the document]
klauer: Agreed - ignoring the step size sign seems appropriate to me 👍

nabs/preprocessors.py
```diff
@@ -301,4 +302,70 @@ def daq_during_plan():
...
+            n = kwargs.pop('num')
+        else:
+            # assumes (det_list, motor, start, stop, num)
+            n = args[-1]
+
+        if not isinstance(n, (int, float)):
+            raise TypeError("Step size / number of steps is "
+                            "neither float nor integer")
+
+        if type(n) is int:
```
klauer, line 342: This is likely excessively cautious, but it's not impossible that the scientists would get step counts/step sizes from numpy calculations which don't return native floats/ints:

An alternative could be, e.g.:

```python
In [3]: isinstance(1, numbers.Integral)
Out[3]: True

In [4]: isinstance(1., numbers.Integral)
Out[4]: False

In [5]: import numpy as np

In [6]: isinstance(np.uint(8), numbers.Integral)
Out[6]: True
```

Otherwise falling back to `step_size = float(n)`

nabs/preprocessors.py
```diff
@@ -301,4 +302,70 @@ def daq_during_plan():
...
+
+    @wraps(plan)
+    def inner(*args, **kwargs):
+        if 'num' in kwargs:
+            # Currently unneeded, since daq_ascan and daq_dscan
+            # do not support num kwarg
+            n = kwargs.pop('num')
+        else:
+            # assumes (det_list, motor, start, stop, num)
+            n = args[-1]
```
klauer, line 336: Rather than assume, should we do:

`det_list, motor, start, stop, n = args`

and fail if it doesn't match up with what's expected?

nabs/preprocessors.py
```diff
@@ -301,4 +302,70 @@ def daq_during_plan():
...
+
+        if not isinstance(n, (int, float)):
+            raise TypeError("Step size / number of steps is "
+                            "neither float nor integer")
+
+        if type(n) is int:
+            # interpret as number of steps (default)
+            yield from plan(*args, **kwargs)
```
klauer, line 344: Do we want to catch the return value of the plan and return it below, just in case?
```suggestion
            result = yield from plan(*args, **kwargs)
```

nabs/preprocessors.py
```diff
@@ -301,4 +302,70 @@ def daq_during_plan():
...
+                                 "provided range and step size.")
+
+            # new endpoint needed, for cases where
+            # (range % step_size) != 0 or to include endpoint
+            if np.isclose(step_list[-1] + n, stop):
+                step_list.append(step_list[-1] + n)
+            n_steps = len(step_list)
+
+            yield from plan(*args[:3], step_list[-1], n_steps,
+                            **kwargs)
```
klauer, line 366: There's no `else:` clause here for contrived errors like `motor, 1, 10, "five"`

nabs/tests/test_plans.py
```diff
@@ -388,3 +388,69 @@ def test_daq_fixed_target_multi_scan(RE, daq, hw, sample_file):
     assert len(reads) == 24
     RE(msgs)
     summarize_plan(msgs)
+
+
+def orange(start, stop, num):
```
klauer, line 393: This is such a nice utility function; why not use it for the scan?

klauer (approved):
Noted some mostly surface-level stuff that could be adjusted, but overall I think this is great. (Honestly, it's much cleaner than whatever I might have come up with 😁 )

nabs/tests/test_plans.py
tangkong: My thought was that I don't need to calculate `orange` for the step number (integer) case, since it can default to the normal `bp.scan` plan.  I could see it being useful to scientists though, maybe.

nabs/preprocessors.py
tangkong: What error does this catch?  I'm probably (definitely) uneducated about this but they look identical to me?

nabs/preprocessors.py
tangkong: I think I catch anything that's not an int or float early.

nabs/preprocessors.py
klauer: Not an error, but the return value of the plan. This ties in with the recent discussion of getting return values from plans with `RE()` rather than a uid. Below, we would return `result` so that we don't just throw away whatever the plan is returning. Or, to be fair, we could just say `return (yield from ...)`, just syntax-wise I personally prefer the above.

The finer points of generators yielding values, sending values, and returning values is admittedly a bit hard to understand, but it's fundamental to how bluesky passes messages/return values back and forth (and does pausing and whatnot).

nabs/preprocessors.py
klauer: My mistake - nice 😁

nabs/preprocessors.py
ZLLentz: Huh, I didn't realize (or forgot?) that numpy datatypes didn't pass basic isinstance checks, though it makes sense.

After reading through the numbers docs: https://docs.python.org/3/library/numbers.html
- All of these numbers are instances of Real (including numpy floats)
- numpy ints and built-in ints are instances of Integral

So the generic "is this an integer" maps to `isinstance(my_number, numbers.Integral)`, and the generic "is this a float" maps to `isinstance(my_number, numbers.Real) and not isinstance(my_number, numbers.Integral)`?

That's pretty terrible if true

ZLLentz: I guess it comes back to the base idea here being really awkward and unintuitive to the uninitiated: there is no reason for programs to differentiate between whether or not the user has given exactly a float in most cases, since we expect 1 and 1.0 to do the same thing in general, so the standard library isn't built around these sorts of use cases.

tangkong:
Any further thoughts on this one?  Do we think it's ready to merge?

klauer:
Nothing further on my end 👍

nabs/preprocessors.py
```diff
@@ -301,4 +303,71 @@ def daq_during_plan():
...
+            det_list, motor, start, stop, n = args
+
+        if not isinstance(n, (numbers.Integral, float)):
+            raise TypeError("Step size / number of steps is "
+                            "neither float nor integer.")
+
+        if isinstance(n, numbers.Integral):
+            # interpret as number of steps (default)
+            result = yield from plan(*args, **kwargs)
+        elif isinstance(n, float):
```
ZLLentz, line 346: I think this type handling and the bit above still needs some small adjustments to handle numpy floats, if that's something that we want to allow

ZLLentz: Other than that I think this is gtg

tangkong: It's a good thing you pushed on this, numpy floats don't behave very nicely around zero, and I've messed with some tolerances to make these work for the test cases I could find.

klauer: It seems like choosing an absolute tolerance value that works for everything could be difficult (imagine working in meters and scanning nanometers or microns, or the same seconds to nanoseconds with the delay stages).

I wonder if something like the following would be an improvement: "add on the `scan end` point if the final calculated point is more than 1/4 of the scan step" (where 1/4 is just a factor I arbitrarily chose). But I think the key _might_ be keeping it relative to the scan step.

Or maybe it's fine as-is and I'm overthinking it.

tangkong: That's an interesting thought.  I think that might result in inconsistent behavior if you approach the same point in large or small steps.  

For what it's worth the current implementation does use a combination of relative and absolute tolerances, they're just defined relative to the scan **points** rather than the step **size**.

klauer: Possibly. That said, I suppose what either of us think here doesn't matter nearly as much as what the original implementation did.

The gold standard is as follows:

https://github.com/pcdshub/old-hutch-configs/blob/200f861c1fb6d66876f1f3817de9a2f24e0245b1/xpp/prod/daq.py#L1262-L1270

tangkong: .... that does solve our problem but I kinda hate it.  

`ascan(0, 2, 0.3) -> [0. , 0.3, 0.6, 0.9, 1.2, 1.5, 1.8, 2.1]`

nabs/utils.py
```diff
@@ -110,3 +112,38 @@ def add_named_kwargs_to_signature(
     )
 
     return sig.replace(parameters=start_params + wrapper_params + end_params)
+
+
+def orange(start, stop, num, rtol=1.e-5, atol=1.e-7):
```
ZLLentz, line 117: It took me a very long time to realize this was probably "o-range" and not the fruit

ZLLentz (approved):
I think we're good to go

nabs/utils.py
tangkong: This method name was my favorite part of the PR, honestly ;D

PR #57, opened 2022-03-22
WIP/ENH: Plan validation framework, extra checks

tangkong:
## Description
Adds a high level `validate_plan` function that runs a plan through multiple checks.  
Adds check for proper plan construction (open_run/close_run matching, proper `run_key` tags)
Adds check for stray calls, pulling from a pre-defined list of forbidden function calls.

## Motivation and Context
Plan verification has always been on our radar, but hasn't been fully fleshed out.  One could see this being useful to scientists building their own plans as a quick check before running them on the beamline.  With the more recently developed RunWrapper functionality, one could see these checks being run before a plan is added to the hutch-python namespace and made available for generic use.

All of these plans attempt to simply examine messages, and as such avoid actually taking any actions. This does not, however, prevent a user from directly moving a motor within a plan via `.put()` or similar methods.

`check_stray_calls` attempts to remedy this by temporarily monkeypatching `sys.modules`.  This may not be the best way to go about it, but it seemed more viable than:
- simple code inspection via AST's, which would not be able to catch calls nested inside helper functions
- static code analysis, which might be a little too heavy

## How Has This Been Tested?
Test suite has been started.  
interactively, for non-simulated motors

Still needs further testing with live PV's, and additional "forbidden methods"

Some docstrings, but mostly this PR

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/56-step_size_decorator.rst
+++ b/docs/source/upcoming_release_notes/56-step_size_decorator.rst
@@ -1,0 +1,22 @@
+56 step size decorator
+#################
+
+API Changes
+-----------
+- 1-Dimensional scans now accept floats in the 'num' argument position and interprets it as a step size.
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- tangkong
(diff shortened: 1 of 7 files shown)


tangkong:
I messed up the commit history, creating new PR based on current master

PR #58, opened 2022-03-22
ENH: Plan validation framework, extra checks

tangkong:
## Description
Adds a high level `validate_plan` function that runs a plan through multiple checks.  
Adds check for proper plan construction (open_run/close_run matching, proper `run_key` tags)
Adds check for stray calls, pulling from a pre-defined list of forbidden function calls.

## Motivation and Context
Plan verification has always been on our radar, but hasn't been fully fleshed out.  One could see this being useful to scientists building their own plans as a quick check before running them on the beamline.  With the more recently developed RunWrapper functionality, one could see these checks being run before a plan is added to the hutch-python namespace and made available for generic use.

All of these plans attempt to simply examine messages, and as such avoid actually taking any actions. This does not, however, prevent a user from directly moving a motor within a plan via `.put()` or similar methods.

`check_stray_calls` attempts to remedy this by temporarily monkeypatching `sys.modules`.  This may not be the best way to go about it, but it seemed more viable than:
- simple code inspection via AST's, which would not be able to catch calls nested inside helper functions
- static code analysis, which might be a little too heavy

It seems that the list of forbidden methods boils down to: `EpicsSignal.put()`
- Stage / unstage methods might only seem 'dangerous' at the point where they try to write to PV, but object information may be written before the `.puts()` are reached.  PV's will not be written to, in either case. 

## How Has This Been Tested?
Test suite has been started.  
interactively, for non-simulated motors

Some docstrings, but mostly this PR

## What needs work?
- [ ] A more complete list of 'forbidden methods'
- [ ] Testing across a variety of devices and live PV's

## Pre-merge checklist


tangkong:
I think this is ready for some eyeballs.  More plan checks and "forbidden methods" are sure to come, but I think there's enough here to start a discussion.

nabs/simulators.py
```diff
@@ -0,0 +1,244 @@
...
+default_patches = [
+            "sys.modules['ophyd'].sim.SynAxis.set",
+            "sys.modules['ophyd'].signal.EpicsSignal.put",
+]
+
+
+@contextmanager
+def patch_sys_modules(
+    modules: List[str] = default_patches
+) -> Generator[Any, None, None]:
```
ZLLentz, line 32: The yield type here is `None`, right?
```suggestion
) -> Generator[None, None, None]:
```

nabs/simulators.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    takes a list of module names as strings and stores them,
+    replaces them with a raiser, and replaces them after
+
+    Need to use exec/eval here due to pass-by-reference issues
```
ZLLentz, line 37: Can you expand a bit on what kind of issues you had? I just want to verify that we need to use these.

I don't immediately see why doing something like
```
from ophyd.signal import EpicsSignal
EpicsSignal.put = raiser
```
would fail to have the same effect.
For that matter, is going through `sys.modules` necessary?

nabs/simulators.py
```diff
@@ -0,0 +1,244 @@
...
+
+def raiser(*args: Any, **kwargs: Any):
+    raise ValidError('forbidden method called')
+
+
+# Be wary of how you specify these, they are keyed based on
+# how they were imported.
+default_patches = [
+            "sys.modules['ophyd'].sim.SynAxis.set",
+            "sys.modules['ophyd'].signal.EpicsSignal.put",
```
ZLLentz, line 25: These seem like reasonable spots to monkeypatch. I was trying to figure out if doing something to the control layer e.g.
```
from ophyd import cl
# do something to the cl here
```
could also work but it seems annoying

nabs/simulators.py
```diff
@@ -0,0 +1,244 @@
...
+        Error message that explains the reason for validation
+        failure. Empty string if validation is successful.
+
+    """
+    success, msg = True, ""
+    try:
+        plan_list = itertools.tee(plan, len(validators))
+        for i, check in enumerate(validators):
+            print(f'running {check.__name__}')
+            check(plan_list[i])
```
ZLLentz, line 191: I hereby declare war on unneeded indexing!
```suggestion
        for plan, check in zip(plan_list, validators):
            print(f'running {check.__name__}')
            check(plan)
```

nabs/simulators.py
```diff
@@ -0,0 +1,244 @@
...
+    print lcls-daq information specifically
+
+    Parameters
+    ----------
+    plan : iterable
+        Must yield `Msg` objects
+    """
+    read_cache: List[str] = []
+    daq_keys = ['events', 'record', 'use_l3t', 'duration']
+    daq_cfg = {k: None for k in daq_keys}
```
ZLLentz, line 216: This misses any stateful nature of the DAQ set up prior to the plan, but I think that's probably OK. Including this information is a bit gross and requires finding the registered daq instance and populating the starting config differently here.

ZLLentz (commented):
I like this a lot! Added some questions/suggestions!

klauer:
This occurred to me the other day; have you done these validations while background threads were running? For example, any ophyd stuff talking to real PVs going on in the background?

In my mind, threading and monkeypatching `sys.modules` seems a recipe for disaster. For the simplicity of this PR, I hope I'm wrong here...

nabs/simulators.py
tangkong: My thought on this was that we're only aiming to summarize what the plan itself does.  If it doesn't configure the DAQ we shouldn't report its configuration and potentially mislead someone into thinking it does.

tangkong:
It's certainly good to question my approach, it seems extreme and I'll certainly be happy if we find a simpler solution. I honestly should have documented my struggles better.

> RE: `from ophyd.signal import EpicsSignal; EpicsSignal.put = raiser`

You're right that this has the same effect from a single session, and it's basically what I'm doing with `sys.modules`. 

I wanted a scheme where we could easily add to the list of modules-to-patch.  Passing in the module itself to a context manager wouldn’t properly patch the method, rather it would just change what the argument references within the function.  If I’m not mistaken, we have to explicitly call something like `EpicsSignal.put = raiser` in the function body to properly patch it

```python
In [11]: from ophyd.sim import SynAxis, motor
    ...: from nabs.simulators import raiser
    ...: 
    ...: 
    ...: def test(module):
    ...:     print(module)
    ...:     print(motor.position)
    ...:     module = raiser
    ...:     print(module)
    ...:     motor.set(1)
    ...:     print(motor.position)
    ...: 
    ...: 
    ...: test(SynAxis.set)
<function SynAxis.set at 0x7fe0fc227280>
0
<function raiser at 0x7fe0fc13f8b0>
1
```

I couldn’t just pass the objects themselves as strings, since they might not be imported explicitly in the namespace the function runs in.
```python
from ophyd.sim import SynAxis
with patch_sys_modules([‘SynAxis.set’]):
    ophyd.sim.motor.set(1)

—> name ‘SynAxis’ is not defined
```
On a more surface level, the modules I wanted access to always seemed to be in `sys.modules` but weren’t always in the namespace of the validation functions.  

I suppose we could modify `patch_sys_modules` to also import the base level device, but this would still involve exec / eval.  

```python
default_patches = [
    ('from ophyd.sim import SynAxis', 'SynAxis.set')
]

@contextmanager
def patch_sys_modules(modules=default_patches)
    cache = {}
    for imp, name in modules:
        try:
            exec(imp)
            cache[name] = eval(name)
            exec(f'{name} = raiser')
…
```

I had stopped looking at the monkeypatch portion of this PR, but looking at it again maybe we can do this with `setattr` instead.  (I don't know why I didn't think about this earlier)

> RE: background threads

That’s a good point, I’m not sure how this would work with threads happening in other processes.  My hope is that this would only patch the `put` methods in the session you’re working on, leaving any screens/subscriptions in other threads alone.  I haven’t checked any of this however. 

I have tested on real PVs, but none that I actually had write-access to.

ZLLentz:
My memory from messing around with this a long time ago is that python "threads" share all the same module resources, while python "processes" (e.g. the multiprocessing module) do not- technically, this means that there are ways to maliciously bypass the monkeypatch.

And even if we e.g. check the thread id before deciding to raise, that makes it possible for someone to write code that does all its epics communications in a background python thread and bypass the raise...

I'm not really sure of any other way to handle these sorts of things at runtime. This is pretty tricky.

ZLLentz:
> A knock-on effect of this is that plans wouldn't be allowed to use multiprocessing inside their threads. But maybe that's ok...

I hope no user is quite this terrifying

klauer:
> Putting the check_stray_calls check into a multiprocess was simple enough

Wow, more succinct than expected! 

Have you run this from a fully populated and talking-to-EPICS hutch-python environment just yet? (e.g., from a normal xcspython session) I'd like to see that pyepics in the subprocess doesn't cause any problems and things along those lines.
So long as that works, I don't think I have any major concerns about this approach.

Then we need to expand this onto other functions, including things like `epics.PV.put` or `epics.caput` and such.

tangkong:
Short update (longer now that I'm done writing it): Within multiprocesses I can monkeypatch the `.set()` call, but objects actually can't `.get()` PV information either.  This is good in that channel access seems to be forbidden, but bad in that reading PV's will also result in a plan failing validation.  

I think grabbing information from PV's should be allowed, even really basic adaptive plans could do this.   I've thought about patching the `.get()` methods too, I'm sure at some point we'd throw another exception if that value is used in calculations mid-plan

On the bright side, this does patch `.set()` methods properly, as long as that’s the only thing that’s being called.

I should have had the guts to poke a real PV's earlier 😞

ZLLentz:
Basic adaptive plans should grab their PV info from `read` calls, right? Like, the bluesky run engine should hand the values off to the plan. I think it's still always improper to `get`

tangkong:
Even if we don't call `get` directly, sometimes `read` will call `get` to update/populate the values.  

https://github.com/bluesky/ophyd/blob/7909d08e28a671eea4f3499568977ebf8c1d4cc3/ophyd/device.py#L1237

ZLLentz:
Maybe I've lost the thread here, I didn't realize we were inspecting the plans by putting them through the run engine? I assumed we were just iterating through the messages and looking for stray set/get outside of the RE mechanics.

tangkong:
You've got the right idea, I just didn't give the full context, sorry.  You are right, we are only looking for stray set/get calls, and no RunEngine is used.  

When I tested this against one of the yag PVs, I ran into a case where the `set` call tried to `get` from a different PV.  Specifically, setting `mec_yag0.focus` eventually tries to get the disabled status

(traceback from within the multiprocess)
```python
File "<ipython-input-29-6b1f0f0626c0>", line 4, in test_plan
    mec_yag0.focus.set(32.)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.1/lib/python3.9/site-packages/ophyd/positioner.py", line 87, in set
    return self.move(new_position, wait=wait, moved_cb=moved_cb,
  File "/cds/home/r/roberttk/devrepos/pcdsdevices/pcdsdevices/epics_motor.py", line 112, in move
    return super().move(position, wait=wait, **kwargs)
  File "/cds/home/r/roberttk/devrepos/pcdsdevices/pcdsdevices/interface.py", line 619, in move
    st = super().move(*args, **kwargs)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.1/lib/python3.9/site-packages/ophyd/utils/epics_pvs.py", line 198, in wrapper
    return fcn(self, *args, **kwargs)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.1/lib/python3.9/site-packages/ophyd/epics_motor.py", line 169, in move
    status = super().move(position, **kwargs)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.1/lib/python3.9/site-packages/ophyd/positioner.py", line 191, in move
    self.check_value(position)
  File "/cds/home/r/roberttk/devrepos/pcdsdevices/pcdsdevices/epics_motor.py", line 499, in check_value
    super().check_value(value)
  File "/cds/home/r/roberttk/devrepos/pcdsdevices/pcdsdevices/epics_motor.py", line 217, in check_value
    if self.disabled.get() == 1:
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.1/lib/python3.9/site-packages/ophyd/signal.py", line 1180, in get
    info = self._get_with_timeout(
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.3.1/lib/python3.9/site-packages/ophyd/signal.py", line 1133, in _get_with_timeout
    raise ReadTimeoutError(
ophyd.signal.ReadTimeoutError: Failed to read HXX:HXM:CLF:01.DISP within 2.00 sec
```

ZLLentz:
Ohh I see, that makes sense now. I guess any of these methods can have epics side-effects.

tangkong:
An update from poking and proddings here and there.  

I wasn't able to find a way to permit channel access through the MultiProcess, and I sort of like it that way.  As an exercise (and at a lack of better solutions), I looked to see what higher-level methods we would need to patch to work with this as is.  I arrived at:

- `SynAxis.set`: Base method for writing fake signals.  For testing
- `EpicsSignal.put`: Base method for writing.  Ideally everything would simply hit this method and we would be done.  This works for the vast majority of cases.  
- `EpicsSignal.set`: put, but for bluesky compatibility.  This actually sometimes starts its own [thread](https://github.com/bluesky/ophyd/blob/f245d9fa76f858ccc3cf5507b7febfe630f43ce8/ophyd/signal.py#L324-L326), which hides exception from our Process.  This can cause `check_stray_calls` to exit successfully before CA times out and throws an error.
- `PositionerBase.set`: A more flexible set call, which can call get methods (eg [check_value](https://github.com/bluesky/ophyd/blob/82053af306815a4980aa4689a31d717f3234bd43/ophyd/positioner.py#L191))

I can't claim to have been exhaustive, but I have looked through some experiment files and these patches cover the common uses of `.set()` and `.put()`.  

I think not supporting PV-reading is the right path, even though it means we can't validate adaptive plans.  Plans that rely on values read dynamically can never be guaranteed to be safe without knowing the hardware state, so applying these plan validators to those is of questionable usefulness.

ZLLentz:
> I think not supporting PV-reading is the right path, even though it means we can't validate adaptive plans. Plans that rely on values read dynamically can never be guaranteed to be safe without knowing the hardware state, so applying these plan validators to those is of questionable usefulness.

I think this is correct. In general, I'd like to see adaptive plans that are aware that they are being iterated outside of the run engine (Dreaming/spitballing a bit: "hmm, the read didn't actually give me anything, so I'm not in a run engine? `yield Msg('AdaptivePlaceholder')`") so that we could audit the rest of the plan successfully without getting some silly `KeyError` or something. But that's out of scope here, surely.

klauer:
> I think not supporting PV-reading is the right path, even though it means we can't validate adaptive plans. 

I'm OK with at least saying "let's not bother with it now". 99% of the plans we support are _not_ adaptive, so ignoring them in the first pass is entirely sufficient.

> In general, I'd like to see adaptive plans that are aware that they are being iterated outside of the run engine

Dreaming about these things is fun. I think adaptive plans could have a decorator (or other mechanism) that provides "test case" parameterized values in order to do a simulation run. I think it would be ideal if those were external to the code itself and didn't require any `if sim_mode/else`-style checks.

tangkong:
In my testing on live PV's, I wasn't able to see any adverse effects from running these functions.  I think the scope of this won't expand any further, so maybe it's time to close this out?

nabs/utils.py
```diff
@@ -114,6 +116,39 @@ def add_named_kwargs_to_signature(
...
+
+    def join_and_raise(self):
+        super().join()
+        # raise exceptions after process is finished
+        if self.exception:
+            raise self.exception[0]
+
+    @property
+    def exception(self):
+        #
```
klauer, line 146:
```
```

nabs/simulators.py
```diff
@@ -0,0 +1,260 @@
...
+
+    """
+    success, msg = True, ""
+    try:
+        plan_list = itertools.tee(plan, len(validators))
+        for plan, check in zip(plan_list, validators):
+            print(f'running {check.__name__}')
+            check(plan)
+    except Exception as ex:
+        print(ex)
```
klauer, line 209: Intentional prints here? What does the validation output actually look like with the current proposed code?

nabs/simulators.py
```diff
@@ -0,0 +1,260 @@
...
+class ValidError(Exception):
+    pass
+
+
+def raiser(*args: Any, **kwargs: Any):
+    raise ValidError('forbidden method called')
+
+
+default_patches = [
+            (SynAxis, 'set'),
```
klauer, line 27: Big indent in this block (2 tabs instead of 1?)

nabs/simulators.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    success, msg = True, ""
+    try:
+        plan_list = itertools.tee(plan, len(validators))
+        for plan, check in zip(plan_list, validators):
+            print(f'running {check.__name__}')
+            check(plan)
+    except Exception as ex:
+        print(ex)
+        msg = (f'Plan validation failed for reason: {str(ex)}')
```
klauer, line 210:
```suggestion
        msg = f'Plan validation failed for reason: {str(ex)}'
```

nabs/utils.py
```diff
@@ -114,6 +116,39 @@ def add_named_kwargs_to_signature(
...
+        self._exception = None
+
+    def run(self):
+        try:
+            super().run()
+            self._cconn.send(None)
+        except Exception as e:
+            tb = traceback.format_exc()
+            self._cconn.send((e, tb))
+            # raise e
```
klauer, line 136:
```suggestion
```

klauer (approved):
A bit of cleanup to do (stray prints, comments, minor style things, etc.) and release notes to add, but I have no further suggestions of real substance for this PR.

Nice work!

tangkong:
@ZLLentz any additional thoughts?  I'm not too bothered about this missing the tag, but it'd be nice to close it out.

ZLLentz (approved):
Fooled myself into thinking I already green checkmarked this

PR #61, opened 2022-05-19
ENH: Read metadata to configure BEC, disable plots for delay_scan

tangkong:
## Description
- adds BECOptionPerRun, which wraps the BEC and looks for run metadata to disable or enable functionality.
- uses the above feature to disable plots for duration_scan, and by extension delay_scan

## Motivation and Context
[Jira](https://jira.slac.stanford.edu/browse/LCLSECSD-1090)

Added to hutch-python in pcdshub/hutch-python#340 

Certain scans give BestEffortCallback some difficulty, and we want a way to turn off plots on a per-run basis.  There aren't any natural (read: easy) ways to do this, so we have to modify the callback a bit to look for metadata tags.  

~~"Normally" one would simply pass a factory function to the RunRouter, but in our case we want to only modify the BEC for a single run.  To reset the BEC, we must remember a bit about its state, leading to this strange FactoryFaker thing.~~

This was done by simply subclassing BEC in a wrapper callback that keeps track of state and reacts to run metadata.

## How Has This Been Tested?
Interactively

Some docstrings, this PR

## Pre-merge checklist

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -22,9 +22,10 @@ requirements:
       - numpy
       - ophyd
       - pandas
-      - pyqt <5.15.0
       - scipy
       - toolz
+    run_constrained:
+      - pyqt <5.15.0
 
 test:
     imports:
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -22,6 +22,7 @@ requirements:
       - numpy
       - ophyd
       - pandas
+      - pyqt <5.15.0
       - scipy
       - toolz
 
--- a/nabs/callbacks.py
+++ b/nabs/callbacks.py
@@ -221,7 +221,7 @@ class BECOptionsPerRun(BestEffortCallback):
         super().stop(doc)
 
     def restore_settings(self):
-        # reset bec settings if they exist
+        """Reset bec settings if they exist."""
         if self.prev_settings:
             for k, v in self.prev_settings.items():
                 setattr(self, k, v)
--- a/nabs/callbacks.py
+++ b/nabs/callbacks.py
@@ -165,3 +165,68 @@ tr:nth-child(even) {
                 tags=['run_table', 'RE'],
                 title='run_table'
             )
+
+
+class BECOptionFactoryFaker():
+    """
+    Takes an instance of the BestEffortCallback, and watches for metadata
+    keys to configure the BEC for a single run
+
+    Fakes a factory function to pass into RunRouter, for the purpose
+    of generating callbacks for a RunEngine.
+
+    Factories need to have the signature: callback(name, start_doc)
+    To restore BEC state after a completed run, we must keep bec on hand
+
+    Usage:
+
+    .. code-block: python
+
+        from event_model import RunRouter
+        from bluesky.callbacks.best_effort import BestEffortCallback
+
+        bec = BestEffortCallback()
+
+        bec_rr = RunRouter([BECOptionFactoryFaker(bec)])
+        RE.subscribe(bec_rr)
+    """
+    # Metadata key : bec attribute
+    valid_keys = {'disable_plots': '_plots_enabled',
+                  'disable_table': '_table_enabled'}
+
+    def __init__(self, bec):
+        self.bec = bec
+        self.prev_settings = {}
+
+    def __call__(self, name, doc):
+        def cb(name, doc):
+            if name == 'start':
+                # store bec settings
+                for k, v in self.valid_keys.items():
+                    if k in doc:
+                        self.prev_settings[v] = getattr(self.bec, v)
+
+                # apply desired changes to bec
+                if 'disable_plots' in doc:
+                    if doc.get('disable_plots'):
+                        self.bec.disable_plots()
+                    else:
+                        self.bec.enable_plots()
+
+                if 'disable_table' in doc:
+                    if doc.get('disable_table'):
+                        self.bec.disable_table()
+                    else:
+                        self.bec.enable_table()
+
+            elif name == 'stop':
+                # reset bec settings
+                for k, v in self.prev_settings.items():
+                    setattr(self.bec, k, v)
+                # clear history
+                self.prev_settings = {}
+
+            # Actually run bec callback
+            self.bec(name, doc)
+
+        return [cb], []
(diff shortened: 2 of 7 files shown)


nabs/callbacks.py
```diff
@@ -165,3 +165,68 @@ def stop(self, doc):
...
+        self.bec = bec
+        self.prev_settings = {}
+
+    def __call__(self, name, doc):
+        def cb(name, doc):
+            if name == 'start':
+                # store bec settings
+                for k, v in self.valid_keys.items():
+                    if k in doc:
+                        self.prev_settings[v] = getattr(self.bec, v)
```
ZLLentz, line 207: Small edge case here: what if the user or the code has some sort of error and we see two start documents in a row without seeing a stop document inbetween? It's possible that we'll lose the true original settings. Maybe we can check if the prev_settings dict has already been cleared before going into this?

nabs/callbacks.py
```diff
@@ -165,3 +165,68 @@ def stop(self, doc):
...
+    of generating callbacks for a RunEngine.
+
+    Factories need to have the signature: callback(name, start_doc)
+    To restore BEC state after a completed run, we must keep bec on hand
+
+    Usage:
+
+    .. code-block: python
+
+        from event_model import RunRouter
```
ZLLentz, line 185: What exactly is this? I must have missed this part of the bluesky ecosystem

nabs/callbacks.py
```diff
@@ -165,3 +165,68 @@ def stop(self, doc):
...
             )
+
+
+class BECOptionFactoryFaker():
+    """
+    Takes an instance of the BestEffortCallback, and watches for metadata
+    keys to configure the BEC for a single run
+
+    Fakes a factory function to pass into RunRouter, for the purpose
+    of generating callbacks for a RunEngine.
```
ZLLentz, line 176: What exactly is being faked here? This looks very real to me

ZLLentz (commented):
I like this, I have a few questions re: the naming and edge cases and things I just don't know about how the internals work

nabs/callbacks.py
tangkong: The RunRouter routes documents through callbacks, basically.  Here it's providing us a layer that reads the document and modify the bec before it sees documents.  You can also use it to modify the documents that callbacks see, leaving the recorded documents intact.  (I've used it in the past to help deal with file handling/saving)

Its docstring says it's an experimental feature, but it's been around for years at this point

nabs/callbacks.py
tangkong: Then normal pattern is to pass a factory function (`Callable[[str, dict], List[Callbacks], List[Subfactories]]`) to the RunRouter, rather than a class.  I guess it's enough like a duck to call a duck though

nabs/callbacks.py
ZLLentz: Okay, I think I understand what the `RunRouter` is, but not why it's used here. We aren't modifying the documents, right? Just reacting to them? What makes this formulation better than the exact same thing, but packaged as a solo thing that can be subscribed to by the RE? Does it save some verbosity + do helpful pre-callback handling?

btw I'm not trying to critique the code, I'm trying to understand for future cases

nabs/callbacks.py
ZLLentz: Gotcha, I see the rationale. I think it would have been ok to just call it a `BECOptionFactory` but it's fine either way.

nabs/callbacks.py
tangkong: hm, no you may be right here.  It might be simpler to just wrap the BEC in another callback that reads the documents.

tangkong: The RunRouter has paths for handling document filling, which is more involved than we need here.  It's probably more intuitive to just subclass BEC and save the state there, which I'll do now

tangkong:
Ugh more pyqt issues

tangkong:
Some pretty significant changes, simplifying things. (thanks Zach for the shrewd eye)

nabs/callbacks.py
```diff
@@ -165,3 +166,60 @@ def stop(self, doc):
...
+    """
+    # Metadata key : bec attribute
+    valid_keys = {'disable_plots': '_plots_enabled',
+                  'disable_table': '_table_enabled'}
+
+    def __init__(self, *args, **kwargs):
+        self.prev_settings = {}
+        super().__init__(*args, **kwargs)
+
+    def start(self, doc):
```
ZLLentz, line 196: This is really clean and should absolutely work as advertised in all normal cases.

Is it all possible to have a glitchy case where:
- We start a run with disable plots
- Something crashes and end_run is never called despite our best efforts in our finalize wrapper constructs
- The plots are still disabled for all scans that follow

ZLLentz: And if that is possible, what's the best way for us to deal with it? I'm not immediately sure.

tangkong: I believe in that case, the user could simply invoke `bec.enable_plots()` and the next time a start doc is seen, that will plots enabled will be the most recent "previous setting"

tangkong: Oh sure I see.  I think we can easily restore settings if they still exist in the start document.  Modification incoming

ZLLentz: Oh I like that solution- simple and clean

ZLLentz (approved):

conda-recipe/meta.yaml
```diff
@@ -22,6 +22,7 @@ requirements:
       - numpy
       - ophyd
       - pandas
+      - pyqt <5.15.0
```
klauer, line 25: Should this be in `run-constrained`? pyqt isn't actually a dependency, I think?
Not a show stopper regardless

nabs/callbacks.py
```diff
@@ -165,3 +166,65 @@ def stop(self, doc):
...
+                self.enable_table()
+
+        super().start(doc)
+
+    def stop(self, doc):
+        self.restore_settings()
+        super().stop(doc)
+
+    def restore_settings(self):
+        # reset bec settings if they exist
```
klauer, line 224:
```suggestion
        """Reset bec settings if they exist."""
```

klauer (approved):
This looks simple and effective to me. Nicely done!

conda-recipe/meta.yaml
tangkong: A good point.  A quick search confirms this.

PR #65, opened 2022-08-15
FIX: daq wrapper duration arg

### LINKED ISSUE #64: BUG: daq scans fail when `duration` is passed in without specifying `events`
## Expected Behavior
DAQ scans should correctly consider both arguments, and respect the defaults.  

## Current Behavior
Reported by @vespos 

```
RE(bp.daq_count([], duration=10, record=False))
...
RuntimeError: Cannot start daq in scan step, did not configure events or duration.
# If I specify a events argument, then it runs fine, but ignores the duration.
```

## Possible Solution
It seems the DAQ interface only works with one of `events` and `duration`, but not both.  The daq scans are passing the arguments to `bps.configure(daq, ... )`, which must house the error.

## Context
Vincent encountered this when some scans failed

## Your Environment
pcds-5.4.1?

ZLLentz:
## Description

- Fix an issue where any of the DAQ step scans would fail if a `duration` arg was passed instead of an `events` arg.
- Make the test suite run on windows for my offline testing.
- Make some of the tests display a better error message to help me track down why they were failing on windows.
- Make the test suite compatible with bluesky v1.9.0

## Motivation and Context

- @vespos pointed this out 2 weeks ago where the scans did not behave as intended. They would error out if we used the `duration` arg. This is used less often so it has gone unnoticed.
- Offline testing is convenient for me.
- closes #64 

## How Has This Been Tested?

I added tests. The test that I added prior to the fix commit fails without the fix. With the fix, the test passes.
I have not run the tests again on linux, hopefully travis can help me here.

The bugfix will be included in the release notes.

## Pre-merge checklist

### DIFF
--- a/nabs/preprocessors.py
+++ b/nabs/preprocessors.py
@@ -112,9 +112,29 @@ def daq_step_scan_wrapper(plan, events=None, duration=None, record=True,
         daq_has_triggered = False
 
     def daq_first_cycle(msg):
-        yield from bps.configure(daq, events=events, duration=duration,
-                                 record=record, use_l3t=use_l3t,
-                                 controls=list(motor_cache))
+        if events is not None:
+            yield from bps.configure(
+                daq,
+                events=events,
+                record=record,
+                use_l3t=use_l3t,
+                controls=list(motor_cache),
+            )
+        elif duration is not None:
+            yield from bps.configure(
+                daq,
+                duration=duration,
+                record=record,
+                use_l3t=use_l3t,
+                controls=list(motor_cache),
+            )
+        else:
+            yield from bps.configure(
+                daq,
+                record=record,
+                use_l3t=use_l3t,
+                controls=list(motor_cache),
+            )
         return (yield from add_daq_trigger(msg))
 
     def add_daq_trigger(msg):
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,7 @@ env:
 
 jobs:
   allow_failures:
-    - name: "Python 3.6 - PIP"
-    - name: "Python 3.7"
-    - name: "Python 3.8"
+    - name: "Python - PIP"
 
 import:
   # This import enables a set of standard python jobs including:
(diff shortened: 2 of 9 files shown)


ZLLentz:
tests fail on travis pip build, but not on my windows laptop or on the lcls linux servers. It seems like `check_limits` requires that a run engine be set up (?), and this appears to not be the case only on travis (?)

ZLLentz:
For some reason I thought it needed pytest-coverage instead of pytest-cov
I'll revert the last commit after we merge https://github.com/pcdshub/pcds-ci-helpers/pull/83

ZLLentz:
3.8 fails on CI because the test suite relies on pcdsdevices which is no longer tested on 3.8 and has 3.9-style annotations.

In principle I'd expect this package to run normally on python 3.8. I'm split between skipping all the pcdsdevices tests and just setting the CI to only check 3.9 and later.

klauer (approved):
The core of the change looks good, and Python 3.9 passing is sufficient to merge this in my opinion. hutch-python/pcdsdevices will only run on 3.9+, so why bother testing with lower versions?

ZLLentz:
Yeah, you're right about that. I put in a silly test skip commit but if it doesn't work out I'll revert and switch to only testing on python 3.9

nabs/preprocessors.py
```diff
@@ -112,9 +112,29 @@ class State:
...
+                daq,
+                duration=duration,
+                record=record,
+                use_l3t=use_l3t,
+                controls=list(motor_cache),
+            )
+        else:
+            yield from bps.configure(
+                daq,
+                record=record,
```
tangkong, line 134: This assumes the daq was configured first, but what if it wasn't?  (the answer is a RuntimeError, as before 👍 )

tangkong (approved):
LGTM, the diff makes it look like you snuck in the daq fix alongside a bunch of test fixes

nabs/preprocessors.py
ZLLentz: That's right, the user will get the previously seen error but it will actually be correct this time!

PR #71, opened 2023-04-20
MAINT: bulk secrets and readme update

ZLLentz:
In all repos:
- Add secrets: inherit to the gh workflow so that the workflow has access to the pypi token on tag
- If the repo has a README.md, ensure that it is packaged correctly in pyproject.toml

### DIFF
--- a/.github/workflows/standard.yml
+++ b/.github/workflows/standard.yml
@@ -10,6 +10,7 @@ on:
 jobs:
   standard:
     uses: pcdshub/pcds-ci-helpers/.github/workflows/python-standard.yml@master
+    secrets: inherit
     with:
       # The workflow needs to know the package name.  This can be determined
       # automatically if the repository name is the same as the import name.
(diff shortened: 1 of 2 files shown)


tangkong (approved):

tangkong:
I'm spending too much time looking into these tests.  The 3.9 conda tests grabs pcdsdevices 7.1.0, for whatever reason.  Not only that, but it picks the earliest build number (#0 out of the 3 builds), the one that doesn't include lightpath as a dependency.  I don't know exactly which package pin causes this, but the other conda build grab 7.3.0 and lightpath just fine 🤷

ZLLentz:
I think it must have some other build problem with the lightpath deps if it wants to avoid lightpath so badly. With enough scrutiny we could probably come up with a precise pin for forcing it.
Starting by requiring pcdsdevices>=7.2.0 in the dev requirements might reveal the issue... maybe I should give it a try in a follow-up PR

Issue #73, opened 2023-06-19
nabs test suite fails with bluesky=v1.11.0


ZLLentz:
Tag details: https://github.com/bluesky/bluesky/releases/tag/v1.11.0
Some test suite excerpts from the "big integration test" at https://github.com/pcdshub/pcds-envs/actions/runs/5307584975

1 test failure relating to extra elog posts:
<details>

```
2023-06-19T04:19:26.5207980Z ______________________________ test_elog_callback ______________________________
2023-06-19T04:19:26.5208497Z 
2023-06-19T04:19:26.5208886Z RE = <bluesky.run_engine.RunEngine object at 0x7fd97ceb24f0>
2023-06-19T04:19:26.5210077Z hw = namespace(motor=SynAxis(prefix='', name='motor', read_attrs=['readback', 'setpoint'], configuration_attrs=['velocity',...ted', read_attrs=['softpos', 'noop', 'noop.readback', 'noop.setpoint'], configuration_attrs=['noop'], concurrent=True))
2023-06-19T04:19:26.5210742Z elog = <nabs.tests.conftest.elog.<locals>.MockELog object at 0x7fd979d837f0>
2023-06-19T04:19:26.5211223Z ipython = <nabs.tests.conftest.ipython.<locals>.MockIPython object at 0x7fd979d83f70>
2023-06-19T04:19:26.5211511Z 
2023-06-19T04:19:26.5211955Z     def test_elog_callback(RE, hw, elog, ipython):
2023-06-19T04:19:26.5212290Z         elogc = ELogPoster(elog, ipython)
2023-06-19T04:19:26.5212579Z         elogc.enable_run_posts = True
2023-06-19T04:19:26.5212835Z     
2023-06-19T04:19:26.5213079Z         elog_uid = RE.subscribe(elogc)
2023-06-19T04:19:26.5213312Z     
2023-06-19T04:19:26.5213652Z         ipython.user_ns["In"].append(
2023-06-19T04:19:26.5213949Z             "RE(bp.scan([hw.det], hw.motor, 0, 1, num=10))"
2023-06-19T04:19:26.5214190Z         )
2023-06-19T04:19:26.5214457Z         RE(bp.scan([hw.det], hw.motor, 0, 1, num=10))
2023-06-19T04:19:26.5214698Z     
2023-06-19T04:19:26.5214943Z         assert len(elog.posts) == 2  # start and table posts
[...]

PR #74, opened 2023-07-24
MNT/BLD: unpin pyqt

klauer:
## Description
Unpin pyqt, leaving it at 5.x.

## Motivation and Context
We're looking forward to Python 3.11 everywhere.

`hutch-python > nabs` dependency means that hutch-python and these libraries currently require `pyqt < 5.15` in our environment. Those versions are not available for Python 3.11. PyDM and our other libraries are now (likely/mostly) compatible with PyQt 5.15, so releasing the pin here is our next step toward that transition.

## How Has This Been Tested?
Untested. Relying on CI, which is likely to fail due to bluesky-related changes.

This PR text.

## Pre-merge checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,6 +24,6 @@
 - [ ] Code contains descriptive docstrings, including context and API
 - [ ] New/changed functions and methods are covered in the test suite where possible
 - [ ] Test suite passes locally
-- [ ] Test suite passes on travis
+- [ ] Test suite passes on GitHub Actions
 - [ ] Ran docs/pre-release-notes.sh and created a pre-release documentation page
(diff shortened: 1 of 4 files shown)


tangkong:
These tests are failing due to the bluesky update I think

klauer:
Should we run-constrain bluesky or update based on the thoughts in #73 ?

tangkong:
I want to remember hearing something about bluesky 1.11 being pulled/reverted, or at least thought about more.  That tips my scales toward pinning, on the off chance some implementation details change.  

It's also less work for us now, at the cost of being confused about who made the pins later 😆

ZLLentz (approved):
The bluesky pin is also appropriate, we can unpin it when we make the preprocessors here compatible with plans that use the new message introduced in 1.11

PR #76, opened 2023-08-09
ENH: raise an error if the daq has not yet been configured

klauer:
## Description
Raise an error if the daq object is not available/has not been instantiated/configured in the current session.

## Motivation and Context
@vespos reported that a simple daq scan raised the following error:

```
>>> RE(bp.daq_ascan([], sim.fast_motor1, 920, 930, 11, events=120, record=False))
...
AttributeError: 'NoneType' object has no attribute 'parent'
```

This indicates to me that `None` is making its way into the detector list in bluesky.
Getting a confusing error like this isn't very friendly, so let's catch it early.

This PR text.

## Pre-merge checklist
- [x] Test suite passes on GitHub Actions

### DIFF
--- a/docs/source/upcoming_release_notes/76-no-daq-raise.rst
+++ b/docs/source/upcoming_release_notes/76-no-daq-raise.rst
@@ -1,0 +1,23 @@
+76 no-daq-raise
+#################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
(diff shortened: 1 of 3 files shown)


tangkong (approved):
This looks good to me on the whole, I was concerned about other apps / code paths that might rely on this, but after some brief poking around I think this is appropriate

ZLLentz (approved):
Same thoughts as in the pcdsdaq-side PR, the original traceback without this was rough to parse

PR #79, opened 2024-06-20
TST/BLD: unpin upper bluesky constraint

tangkong:
## Description
Try releasing the upper pin on bluesky for future illumine compatibility

## Motivation and Context
Revisiting this now that numpy 2.0 turned our lives upside down

## How Has This Been Tested?
Go gha go

This PR

## Pre-merge checklist
- [x] Test suite passes on GitHub Actions

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -21,7 +21,7 @@ requirements:
   - pip
   run:
   - python >=3.6
-  - bluesky >=1.6.5,<1.11
+  - bluesky >=1.6.5
   - numpy
   - ophyd
   - pandas
--- a/docs/source/upcoming_release_notes/79-bld_unpin_bluesky.rst
+++ b/docs/source/upcoming_release_notes/79-bld_unpin_bluesky.rst
@@ -1,0 +1,22 @@
+79 bld_unpin_bluesky
+####################
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- Updates test suite and some plans for numpy 2.0 compatibility (primarily with updated bluesky >1.12.0)
+
+Contributors
+------------
+- tangkong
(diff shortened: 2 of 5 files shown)


tangkong:
bluesky only just updated for numpy 2.0 compatibility, there's a release notes PR up to tag those changes in v1.13.0a4.

tangkong:
after some stupidity reporoducing this locally, a few notes:
* I've confirmed that latest bluesky removes all the numpy deprecation errors.  This version (1.13.0a4?) is not released yet
* Local test errors remaining are:
  * Recursion depth in test_optimize
  * unpickable _thread.lock
  * trying to call `.send()` on a list
* Recursion depth errors come from bluesky trying to deepcopy a `SynGauss` object that was stashed in the md-dictionary, not from the optimization routine itself

...investigation ongoing

tangkong:
The `delay_scan` error I've encountered locally does not seem to appear in CI.  Currently CI only fails when trying to call `pack_event_page` with no arguments.  These go away once we upgrade past 1.12.0 (as can be seen in the conda builds)

I think this PR has correct changes, and the rest of the nabs test suite will have to wait until bluesky updates to 1.13.0a4

conda-recipe/meta.yaml
```diff
@@ -21,7 +21,7 @@ requirements:
   - pip
   run:
   - python >=3.6
-  - bluesky >=1.6.5,<1.11
+  - bluesky >=1.6.5
```
ZLLentz, line 24: Does this need to be re-pinned higher? Or maybe this is for the follow-up?

ZLLentz (approved):
I think these are all correct too

conda-recipe/meta.yaml
tangkong: I'll leave this for the followup.  Older versions of bluesky would still work for numpy<2.0, technically.

tangkong:
Though it hurts to bypass the tests, I'll ship this and we'll wait for the bluesky pin

